PR #34, opened 2019-07-15
Add a 'migrateable' flag on a component

technosophos:
See comments on #12.

I am not totally sure this is ready for merge. We might need to describe a little more fully what we mean by `migrateable`. In this case, I just indicated that if this flag is `false`, the workload(s) attached to this component instance must be destroyed and recreated.


technosophos:
Closing. Seems unnecessary right now

PR #36, opened 2019-07-15
Add 'private' flag to parameter.

### LINKED ISSUE #29: Parameter should have and additional "secret" type
In [Component Model - Parameters](https://github.com/microsoft/hydra-spec/blob/master/3.component_model.md#parameter), I think it would be beneficial to add an additional `Secret` type (compare to `securestring` type in ARM template (https://docs.microsoft.com/en-us/azure/azure-resource-manager/template-best-practices#security-recommendations-for-parameters)).

Secret parameters can't be read after deployment and MUST be redacted from all logs.

technosophos:
This would make it possible for user agents to redact values taht should not be shown in places like log files or user dialogs.

Closes #29


technosophos:
Closing. I don't think we'll be adding this in the near future.

Issue #182, opened 2019-10-03
Resources is optional, but if specified, requires CPU and memory


technosophos:
https://github.com/microsoft/hydra-spec/blob/master/3.component_model.md#resources

Resources is an optional field on a container. But if it is specified, then CPU and Memory are required. This doesn't make sense. CPU and Memory should have sensible defaults and not be required.

technosophos:
Okay, it looks like the Scylla/Rudr implementation is incorrect. `resources` should have been required. I will file a bug there and fix. But I do think that this is the wrong approach; I think resource limits should be optional, and the platform itself should be responsible for sensible defaults.

PR #217, opened 2019-10-15
be explicit about ports for workers and servers

### LINKED ISSUE #199: If a Server or SingletonServer does not have a port on the container, is this an error?
See this issue in Scylla for context.

https://github.com/microsoft/scylla/issues/323

Essentially, the spec does not require a port for Server, but without a port, isn't it just a Worker?

technosophos:
Closes #199

Signed-off-by: Matt Butcher <matt.butcher@microsoft.com>

### DIFF
--- a/3.component_model.md
+++ b/3.component_model.md
@@ -266,15 +266,13 @@ Resources describe compute resources attached to a runtime.
 
 | Attribute | Type | Required | Default Value | Description |
 |-----------|------|----------|---------------|-------------|
-| `cpu` | [`CPU`](#cpu) | N |  | Specifies the attributes of the cpu resource required for the container. |
-| `memory` | [`Memory`](#memory) | N | | Specifies the attributes of the memory resource required for the container. |
+| `cpu` | [`CPU`](#cpu) | Y |  | Specifies the attributes of the cpu resource required for the container. |
+| `memory` | [`Memory`](#memory) | Y | | Specifies the attributes of the memory resource required for the container. |
 | `gpu` | [`GPU`](#gpu) | N |  | Specifies the attributes of the gpu resources required for the container. |
 | `volumes` | [`[]Volume`](#volume) | N | | Specifies the attributes of the volumes that the container uses. |
 | `extended` | [`[]ExtendedResource`](#extendedresource) | N | | Implementation-specific extended resource requirements |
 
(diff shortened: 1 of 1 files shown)


technosophos:
Closes #182

3.component_model.md
```diff
@@ -258,14 +266,16 @@ Resources describe compute resources attached to a runtime.
...
+| `cpu` | [`CPU`](#cpu) | N |  | Specifies the attributes of the cpu resource required for the container. |
+| `memory` | [`Memory`](#memory) | N | | Specifies the attributes of the memory resource required for the container. |
 | `gpu` | [`GPU`](#gpu) | N |  | Specifies the attributes of the gpu resources required for the container. |
 | `volumes` | [`[]Volume`](#volume) | N | | Specifies the attributes of the volumes that the container uses. |
 | `extended` | [`[]ExtendedResource`](#extendedresource) | N | | Implementation-specific extended resource requirements |
 
 For any resource that cannot be satisfied by the underlying platform, the platform MUST return an error and cease deployment. A resource is considered a requirement, and failure to meet that requirement means the runtime MUST NOT deploy the application. For example, if an application requests `1P` of memory, and that amount of memory is not available, the application deployment must fail. Likewise, if an application requires `1` gpu, and the runtime does not provide gpus, the application deployment MUST fail.
 
+If resources are not specified, the underlying platform is free to choose a suitable default.
```
BharatNarasimman: I think this might cause weird & hard to debug errors. Is there a specific reason for making this to be not required?

technosophos: Yes -- issue #182

technosophos: And this was complicated by the fact that the way different platforms measure CPU and memory turn out to be different, as we discovered on Kubernetes a few weeks back.

technosophos: Or maybe I should ask. In Kubernetes, a sensible default for CPU is somewhere around `0.05` and maybe `0.1`. I am assuming that for SF, the sane default is `1`.

BharatNarasimman: We can do fractional cores as well, but I am thinking of cases where the default causes activation of the container to fail and makes it hard and not obvious for the operator to debug(we have seen such cases when we were doing SF Mesh).
Also, the issue #182 mentions Resources are optional for containers, they are mandatory as well in the current schema(is that comment meaning different resources than this one below?). 
resources | Resources | Y |   | Resources required by the container.-- | -- | -- | -- | --

technosophos: Removed that section from this PR, opening another PR to deal with just #182

PR #221, opened 2019-10-16
adding new how it works asset

rbitia:
Changed the color scheme of the asset to be purple.


rbitia:
Signed!

technosophos:
Ronan is checking on the color palette. Do not merge until we get his OK.

vturecek:
Looks nice! But I do worry it wouldn't match the color palette we have on the landing site and overall for the project. If we feel a color change is appropriate, we should open up an issue to track all the things that would need to change. Personally I like the colors we have now.

cc @flynnduism

flynnduism:
Agree - right now the overall OAM theme is primarily blue (with some light green and purple accents) - ideally we should keep things coordinated. If there's an issue with the current direction then we should discuss it and address it broadly across diagrams / logos / website.

Screen Shot 2019-10-15 at 8 02 40 PM

_current oam specification / app ‘blueprints' theme_

rbitia:
Nope I'm cool with blue if it was already coordinated. Maybe we could add the color scheme to the assets so folks know. 

Closing for now.

PR #223, opened 2019-10-16
readme edit

rbitia:
Focusing on OSS projects

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,6 @@
 # Open Application Model Specification
 
-From the creators of 
-[Helm](https://helm.sh),
-[OpenKruise](https://openkruise.io/en-us/),
-the Open Application Model is a specification for building cloud native applications.
+Open Application Model is a specification for building cloud native applications.
 
 Focused on separating concerns of development and operation needs, _Open Application Model_ brings modular, extensible, and portable design to building and delivering applications on platforms like Kubernetes.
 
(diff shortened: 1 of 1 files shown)


technosophos:
Why are we removing this? They did actually contribute to the spec, and Service Fabric is an open source project.

rbitia:
I agree but I don't think that's something folks reading the repo should first see or get held up on. Maybe we can move this line to another file?

vturecek:
Hm, seems weird to not mention other projects that contributors have worked on just because those projects don't meet some standard of open source (SF technically is open source, even though it's not developed openly). I think SF is very relevant to the distributed applications space of this spec, open source or not. I don't think we should merge this change.

rbitia:
I made a fix can you go off that? I don't think you need to mention who the creators are in the first paragraph. I think they are better suited for the collaboration section.

resouer:
Ah, this is interesting. I just got a feedback immediately after we shared the link to one partner that this line "is not community friendly". I think the main reason is we list it as the first line of the README of spec.

I will +1 for Ria.

technosophos:
I am okay with moving to the bottom, or even deleting the entire sentence.

hongchaodeng:
Well, a hero mention little about his past glories.

Let's delete the line and focus on the making OAM the next great community project :)

technosophos:
@rbitia Would you be okay just removing that entire sentence? I think we're all starting to question its value now.

rbitia:
Yeah I'm cool with that too! I can push up another fix

suhuruli:
I am cool with this change (Y)

vturecek:
Removing it entirely sounds better to me than just removing one.

PR #230, opened 2019-10-24
add link to google calendar

suhuruli:


suhuruli:
@tomkerkhove , this should be fixed now. Let me know!

tomkerkhove:
Works fine, thanks!

vturecek:
Is this a general calendar for our google group, or is the link just adding an event to my google calendar?

vturecek:
OK, seems fine, but is there a way to create a shared calendar for the group?

suhuruli:
What do you mean shared calendar? What this does is it adds the OAM Community Calendar to a user's Google account so they can see when OAM community events are held. I suggest we merge this small change and take the discussions to Gitter.

PR #236, opened 2019-10-29
add stage for trait

### LINKED ISSUE #235: How Traits can be composed in order？
Now in rudr, we have implementation to manage whether the trait manager need to execute trait before Component create/modify/delete or after.

So I suggest we add field like `stage` in Trait to define the compose order.

```
kind: Trait
metadata:
  name: Ingress
  annotations:
    version: v1.0.0
spec:
  stage: afterStart/afterModify
  appliesTo:
    - Server
  parameters:
     ...
```

```
kind: Trait
metadata:
  name: Autoscaler
  annotations:
    version: v1.0.0
spec:
  stage: preStart/preModify
  appliesTo:
    - Server
  parameters:
     ...
```

wonderflow:
fixes #235 

When we add some traits, maybe these traits are external resources, some external resource should be provisioned before workload create, some external resource should be bind to workload so they should execute after workload created.

In this case, we add `stage` to define order


wonderflow:
we discussed at #235  and finally result to leave this to runtime implementation

PR #241, opened 2019-11-01
scope: fix Network to allow multiple subnet IDs

hongchaodeng:
This is a recent finding when we are applying OAM in practice:

There could be more than one subnet because there is a limit in the number of IPs in a subnet.
If IPs are taken up, operators need to add another subnet into this network.


wonderflow:
Maybe type `[]string` for `subnet-ids` is better

hongchaodeng:
@wonderflow No definitions for `[]string` exists now.

PR #259, opened 2019-11-15
add description and required for workloadsetting

wonderflow:
Now WorkloadSetting is used to describe what kind of parameters can be used for extended workload in ComponentSchematic.

It's necessary to give description and whether it's required in WorkloadSettings.


wonderflow:
Closing as we are promoting WorkloadSetting as user defined schema

PR #263, opened 2019-11-19
declare that OAM runtime could have more extended application scopes

### LINKED ISSUE #254: Allow runtime-specific extended Application Scopes?
Hi, 

The Application Scope spec mentions that any user defined scopes are not supported atm. Any plans to support this? 

I'm looking if its possible to plug in a different implementation for a custom network scope, where this implementation would be entirely different from implementation of core.oam.dev/v1alpha1.Network.   From my understanding, any implementation for core.oam.dev/v1alpha1.Network should conform to the existing spec where it takes only the parameters network-id, subnet-ids and internet-gateway-type.

wonderflow:
fixes #254 

We should declare that OAM runtime could define more application scopes then what defined in the spec now.

### DIFF
--- a/4.application_scopes.md
+++ b/4.application_scopes.md
@@ -41,7 +41,7 @@ Core application scope types define grouping constructs for basic runtime behavi
 
 ### Extended application scope types
 
-Extended application scopes are optional per runtime, meaning that each runtime may choose which extended scopes are supported. In this version of the spec, allowing user-defined extended application scope types is not supported.
+Extended application scopes are optional per runtime, meaning that each runtime may choose which extended scopes are supported. It is expected that runtimes will implement extended scopes as they need. In this version of the spec, allowing user-defined extended application scope types is not supported.
 
 ## Defining an application scope
 This section is normative because application scopes are an inspectable (and possibly shareable) part of the system. All scopes MUST be representable in the following format.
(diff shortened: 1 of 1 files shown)


4.application_scopes.md
```diff
@@ -187,10 +187,13 @@ spec:
 ```
 
 ## Extended application scope type definitions
+
 The following extended scope types are available:
 - resource quota scope
 - identity scope
 
+These two are just definition examples, the runtime could have more extended application scopes as they need.
```
hongchaodeng:
```suggestion
Besides the above scopes, the runtime could have more extended application scopes as needed.
```

vturecek:
I have no problem with the change you made, @wonderflow, because _of course_ runtimes should be able to define any scopes they want, just like they can with traits. In fact we could do a bit more work here to establish that is _expected_ that runtimes will implement extended scopes, just like we do with traits.

#254 was initially aimed at loosening this statement on line 44 of application_scopes.md: "In this version of the spec, allowing user-defined extended application scope types is not supported."

wonderflow:
@vturecek  After discussion in that issue, we have figured out what they really want is runtime extended scopes. So we just need to declare clearly here.

Let me add this in.
> it is expected that runtimes will implement extended scopes, just like we do with traits.

vturecek:
OK, the change here LGTM as is, but I do think there is still more work to do here. I'll go ahead and merge this one.

PR #260, opened 2019-11-18
setup i18n framework

wonderflow:
Now I have setup a framework for Chinese Translation. I think we could work together and split the work.

\cc @KaiwenWang @captainroy-hy @resouer 

refs #258 #233

PR #274, opened 2019-11-27
setup i18n framework

wonderflow:
This is same PR with https://github.com/oam-dev/spec/pull/260

Move to i18n branch.


wonderflow:
@vturecek @hongchaodeng  please take a look

PR #275, opened 2019-11-29
Remove "in MB" from memory description

### LINKED ISSUE #271: Description in memory "in MB" should remove
# Expect
The default unit of memory should be bytes, eg: 1024 is 1024 bytes. The description in memory should remove "in MB"

> The minimum amount of memory in MB required for running this container. The value should be a positive integer, greater than zero.

I suggest change to this
> The minimum amount of memory required for running this container. The value should be a positive integer or positive integer using one of these suffixes: P, T, G, M, K.

# Current

https://github.com/oam-dev/spec/blob/master/3.component_model.md#memory

### Memory
Attribute | Type | Required | Default Value | Description
-- | -- | -- | -- | --
required | string | Y |   | The minimum amount of memory in MB required for running this container. The value should be a positive integer, greater than zero.

Memory and disk space use the notation for bytes/kilo/mega/giga/tera/peta by just using the major unit:

1024 is 1024 bytes
88K is 88 kilobytes
5M is 5 megabytes
7G is 7 gigabytes
100T is 100 terabytes
9999P is 9999 petabytes

fatkun:
Fixed #271 

The default unit "MB" in memory description is conflict with unit description.

### DIFF
--- a/3.component_model.md
+++ b/3.component_model.md
@@ -286,7 +286,7 @@ See the Units section above for valid values.
 
 | Attribute | Type | Required | Default Value | Description |
 |-----------|------|----------|---------------|-------------|
-| `required` | `string` | Y | | The minimum amount of memory in MB required  for running this container. The value should be a positive integer, greater than zero. |
+| `required` | `string` | Y | | The minimum amount of memory required  for running this container. The value should be a positive integer or positive integer using one of these suffixes: P, T, G, M, K. |
 
 See the Units section above for valid values.
 
--- a/3.component_model.md
+++ b/3.component_model.md
@@ -286,7 +286,7 @@ See the Units section above for valid values.
 
 | Attribute | Type | Required | Default Value | Description |
 |-----------|------|----------|---------------|-------------|
-| `required` | `string` | Y | | The minimum amount of memory required  for running this container. The value should be a positive integer or positive integer using one of these suffixes: P, T, G, M, K. |
+| `required` | `string` | Y | | The minimum amount of memory required for running this container. The value should be a positive integer with/without unit suffix: P, T, G, M, K. If no unit is given, it defaults to 'bytes'. |
 
 See the Units section above for valid values.
 


vturecek:
@fatkun good catch, this makes more sense. 

What if users omit the units and just provide a number? We can either make the units required by saying a runtime MUST throw an error if units are not provided, or we can use a default unit. I'm in favor of making units required.

fatkun:
> @fatkun good catch, this makes more sense.
> 
> What if users omit the units and just provide a number? We can either make the units required by saying a runtime MUST throw an error if units are not provided, or we can use a default unit. I'm in favor of making units required.

I prefer use default unit (bytes). It's same with k8s spec.

3.component_model.md
```diff
@@ -286,7 +286,7 @@ See the Units section above for valid values.
 
 | Attribute | Type | Required | Default Value | Description |
 |-----------|------|----------|---------------|-------------|
-| `required` | `string` | Y | | The minimum amount of memory in MB required  for running this container. The value should be a positive integer, greater than zero. |
+| `required` | `string` | Y | | The minimum amount of memory required  for running this container. The value should be a positive integer or positive integer using one of these suffixes: P, T, G, M, K. |
```
hongchaodeng: Can you add a line saying the default would be "bytes" if user dismiss any unit?

fatkun: what about your options?
1. The minimum amount of memory required in bytes for running this container. The value should be a positive integer or positive integer using one of these suffixes: P, T, G, M, K.

2. The minimum amount of memory required for running this container. The value should be a positive integer or positive integer using one of these suffixes: P, T, G, M, K. The default unit is bytes if no suffix.

hongchaodeng: second one modified:
```
The minimum amount of memory required for running this container. The value should be a positive integer with/without unit suffix: P, T, G, M, K. If no unit is given, it defaults to 'bytes'.
```

fatkun: done

fatkun:
done @hongchaodeng

hongchaodeng:
lgtm

We would need another maintainer's lgtm before merging

wonderflow:
LGTM

PR #276, opened 2019-12-02
fix definition of application scope to be properties

wonderflow:
When our cloud team try to implement a kind of scope, we find the scope is lack of extension as it can only support `string`, `number` or `bool`.

But I find @technosophos  and @vturecek  have discussed this issue (https://github.com/oam-dev/spec/pull/7) and made a decision to use properties. 

So I think this is just the missing part of that PR.


wonderflow:
closing as we have moved to v1alpha2 and Scope was created by standalone CRD now

Issue #272, opened 2019-11-25
Proposal: OAM for CRD + Operator, make workload have more extension ability like trait 


wonderflow:
Now we could clearly understand Component and Trait, they both have extensions. The extension for OAM in K8s world is obviously using CRD+Operator. 

We could define any CRD+Operator as a Trait by using json schema. However, we can't do the same thing for Workload.

Trait has properties which allow us to define json schema.  While we can only define key-value pairs for workload.That is why some users complain we are not friendly to extended workload. 

So I propose we could let `Workload Types` use jsonschema to define like trait. Then, we could extend those existing CRD+Operators (eg. OpenFaaS) more easily into `Workload Types`.

On the other hand, we should change [the type of workloadsettings](https://github.com/oam-dev/spec/blob/master/3.component_model.md#spec) to be [`Properties`](https://github.com/oam-dev/spec/blob/master/6.application_configuration.md#properties).

What's more, we could have a OAM registry in the future. Users could register their CRD to this registry, if the CRD is a kind of workload, they will get an automatically generated `Workload`; if the  CRD is a kind of operator trait, they will get an automatically generated `Trait`.

\cc @resouer @hongchaodeng

wonderflow:
what about your ideas? @technosophos @vturecek

vturecek:
This is closely related to #268. My response there was basically that the core workload type concept might be too closely tied to containers, when there are other workloads (like processes) that could be supported by those core workload types.

Using a JSON schema properties object instead of the current workloadSettings is a step in the right direction, but I think we may also need to explore decoupling core workload types from the container spec. 

For example, could I define an extended workload type (like processes) that can be used with the core workload types (server, worker, task)?

wonderflow:
> For example, could I define an extended workload type (like processes) that can be used with the core workload types (server, worker, task)?

What if let the workloadsettings be also used by core workload type?

hongchaodeng:
Per https://github.com/oam-dev/spec/issues/281#issuecomment-565271085,
Let's close this and converge discussion to https://github.com/oam-dev/spec/issues/281

PR #277, opened 2019-12-02
Change Component to be an instance, workload Type as schema

### LINKED ISSUE #281: [Proposal] Change Component to be an instance, not schema
# Problem description

## Problem 1: Upgrading Application requires writing another Component

Currently the container image is hard-coded in ComponentSchematic. Each time a new version of the application needs to rollout, we need to write a new ComponentSchematic and update ApplicationConfiguration simultaneously. This combination of work is complex. We heard complains from our users often, and has raised issue https://github.com/oam-dev/spec/issues/265. We want to let developers define only one component and update image in one place.

## Problem 2: Mapping existing Application definition to OAM is not intuitive

Many existing projects (both open source and internally) have defined _Application_ schema and done deployment with instances of _Application_. For examples:

- Crossplane defines _KubernetesApplication_ CRD and let users deploys  _KubernetesApplication_ CR.
- Alibaba: It's similar to what Pinterest did in [this blog](https://medium.com/pinterest-engineering/building-a-kubernetes-platform-at-pinterest-fb3d9571c948). Our projects define AlibabaService CRD. Alibaba's internal projects define AlibabaService CRD. Developers write AlibabaService CR yaml and put it in Git, which then triggers CI/CD pipeline.

The current model having Component as another schema on top of Workload as a schema doesn't fit well for these real world cases.

# Proposal

There are two issues in OAM that we want to improve:  
  
1. Change Component to be an _instance_ instead of _schema_, and 
2. Change workload settings to _schema_, and move all of workload related information here, e.g. container definition, from Component.

## Walkthrough

Let me give a concrete example. Let's say I have defined a WorkloadType AlibabaService:

```yaml
apiVersion: oam.dev/v1alpha1
kind: WorkloadType
metadata:
  name: alibaba-service
spec:
  group: alibaba.io/v1
  names: [AlibabaService]
  settings: |
    {
      "$schema": "http://json-schema.org/draft-07/schema#",
      "type": "object",
      "required": [
        "image"
      ],
      "properties": {
        "image": {
          "type": "string",
          "description": "The container image to deploy"
        }
      }
    }
```

Then deploy a Component:

```yaml
apiVersion: oam.dev/v1alpha1
kind: Component
metadata:
  name: my-nginx
spec:
  workloadType: alibaba.io/v1.AlibabaService
  workloadSettings:
    image: nginx/v1
```

While it might look the same up until now, the key change happens when we write ApplicationConfiguration:

```yaml
spec:
  Components:
  - componentRef: my-nginx
    traits:
    - name: ingress
    - name: monitoring
```

As described above, Component becomes an instance itself, and ApplicationConfiguration references instances of Components.

What's changed here is that previously we asked developers how to describe the app, but now we ask them what to deploy. This brings us the following benefits:

- Describing how to run the app is now entirely encapsulated in Workload provided by the infra, and dev will just use the capabilities to deploy the app.
- Dev will now know exactly what the deployment of the app looks like because they write it, reducing the complexities of passing information between dev and ops.
- Ops will focus on attaching operational traits to components.

This better resembles the use cases we have been encountering and solves those problems that users asked.

## How it solves above problems?

Problem 1: Since component is an instance now, updating image would trigger correlated deployments to be updated as well.

Problem 2: Application CRD could be modeled as WorkloadType and Application CR could be modeled as Component. This would intuitively map existing use cases to OAM.

wonderflow:
fixes https://github.com/oam-dev/spec/issues/272
fixes #281

### DIFF
--- a/3.component_model.md
+++ b/3.component_model.md
@@ -218,7 +218,7 @@ While it is not required, it is RECOMMENDED that `image` names be suffixed with 
 
 ### Properties
 
-A properties object (for workloadSettings, trait and scope configuration) is an object whose structure is determined by the workload type, trait or scope property schema. It may be a simple value, or it may be a complex object.
+A properties is an object whose structure is determined by the workloadSettings schema. It may be a simple value, or a complex object.
 
 Properties are validated against the schema appropriate for the workload type, trait or scope.
 
--- a/3.component_model.md
+++ b/3.component_model.md
@@ -63,7 +63,7 @@ The spec defines the constituent parts of a component.
 | `osType` | `string` | N |  | The OS required to host (all of) the component's containers (since containers share a kernel with the underlying host). Possible values include:<ul><li>linux</li><li>windows</li></ul> For extended runtimes, this is passed in unaltered. Default can be none and let the runtime decide where to place the component. |
 | `arch` | `string` | N |  | The CPU architecture required to host (all of) the component's containers (since containers share physical hardware with the underlying host). Possible values include:<ul><li>i386</li><li>amd64</li><li>arm</li><li>arm64</li></ul> Default can be none and let the runtime chose architecture. |
 | `containers` | [`[]Container`](#container) | N | | The OCI container(s) that implement the component. |
-| `workloadSettings` | [`[]WorkloadSetting`](#workloadsetting) | N | | Declaration of non-container settings that should be passed to the workload runtime|
+| `workloadSettings` | [`Properties`](#properties) | N | | Declaration of non-container settings that should be passed to the workload runtime |
 
 All core workload types require `containers` and MUST return an error and cease processing if the `containers` section is empty. However, extended workload types may not require containers. Thus this field is marked optional.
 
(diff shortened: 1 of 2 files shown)


7.workload_types.md
```diff
@@ -45,19 +48,55 @@ This section describes the different naming references for this kind.
...
-| `name` | `string` | Y | | The setting's name. Must be unique per workload type. |
-| `description` | `string` | N | | A description of the setting. |
-| `type` | `string` | Y | | The setting's type (One of `string`, `boolean`, `number`). |
-| `required` | `boolean` | N |`false` | Whether a value _must_ be provided for the setting. |
-| `default` | type indicated by `type` field | N | | The setting's default value. |
-
-The `name` field allows Unicode letters, numbers and `_` and `-`. The `description` field allows those characters plus whitespace and punctuation characters.
+### Properties
+
+A properties object is an object whose structure is determined by the workload type workloadSettings schema. It may be a simple value, or it may be a complex object.
```
hongchaodeng:
```suggestion
A properties is an object whose structure is determined by the workloadSettings schema. It may be a simple value, or a complex object.
```

3.component_model.md
```diff
@@ -216,16 +216,17 @@ While it is not required, it is RECOMMENDED that `image` names be suffixed with
 
 > The name field is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.
 
-### WorkloadSetting
+### Properties
 
-This section describes additional configuration for a workload (settings that are not about a container runtime).
+A properties object (for workloadSettings, trait and scope configuration) is an object whose structure is determined by the workload type, trait or scope property schema. It may be a simple value, or it may be a complex object.
```
hongchaodeng: Could we combine this with the section in  _7.workload_types.md_?

wonderflow: I changed the section in `7.workload_types.md`

7.workload_types.md
```diff
@@ -45,19 +55,46 @@ This section describes the different naming references for this kind.
...
+          "description": "the name to the function"
+        },
+        "image": {
+          "type": "string",
+          "description": "the docker image of the function"
+        }
+      }
+    }
+    
+```
```
hongchaodeng: Also add a AppConfig example?

wonderflow:
closing as v1alpha2 is released

PR #280, opened 2019-12-04
add group/version/kind support trait as crd

### LINKED ISSUE #279: Proposal: OAM for CRD + Operator, Trait should have Group/Version/Kind defined in Trait Spec
We hope we could write CRD+Operator for Trait, related with #272 

Now we could define trait schema using jsonschema in `properties`, the missing part is we should add  Group/Version/Kind definition into the Trait, using some field like `type`.

If we don't have a G/V/K in trait, we can't register a CRD+Operator as a Trait.

wonderflow:
fixes #279

### DIFF
--- a/5.traits.md
+++ b/5.traits.md
@@ -72,8 +72,25 @@ The specification defines two major things: The list of workload types to which 
 
 | Attribute | Type | Required | Default Value | Description |
 |-----------|------|----------|---------------|-------------|
+| `group` | `string` | Y | | The group that this trait belongs to (e.g. `core.oam.dev`)
+| `version` | `string` | Y | | The version of the trait API (e.g. `v1alpha1`)
+| `names` | [`Names`](#Names) | Y | | Information about the names associated with this trait |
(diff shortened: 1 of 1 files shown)


5.traits.md
```diff
@@ -72,9 +72,26 @@ The specification defines two major things: The list of workload types to which
...
+
+This section describes the different naming references for this kind.
+
+| Attribute | Type | Required | Default Value | Description |
+|-----------|------|----------|---------------|-------------|
+| `kind` | `string` | Y | | The proper reference name of the trait (e.g., `AutoScaler`) |
+| `singular` | `string` | N | | The singular human-readable name of this type (e.g., `server`) |
+| `plural` | `string` | N | | The plural human-readable name of this type (e.g., `servers`) |
+
+`kind` must begin with a capital letter, and contain only characters 0030-0039, 0041-005a, and 0061-007a from the Unicode basic Latin characters.
```
artursouza: Is this a project decision to use Unicode basic Latin. Why not just UTF-8?

http://utf8everywhere.org/

wonderflow: I just align it with Workload Type https://github.com/oam-dev/spec/blob/master/7.workload_types.md#names

wonderflow:
since we are working  on v1alpha2 which include this work, I will close it.  ref: https://github.com/oam-dev/spec/pull/319

PR #283, opened 2019-12-17
add a tip to tell one of components or scopes is required

wonderflow:
The field `spec` in AppConfig is required while the fields in `spec` are all optional, this makes confusion, so I add a tip to make it clear


wonderflow:
closing as creating  ApplicationScope was moved out from ApplicationConfiguration

PR #290, opened 2019-12-24
Branch 0.3.0-dev: add group/version/kind support trait as crd 

wonderflow:
same PR with #280 that merges to 0.3.0-dev first


hongchaodeng:
lgtm

PR #292, opened 2019-12-25
add expose/consume field for data exchange between components

wonderflow:


hongchaodeng:
I just rebase 0.3.0-dev branch from master and push.

Please rebase against latest 0.3.0-dev branch.

3.component_model.md
```diff
@@ -64,6 +64,9 @@ The spec defines the constituent parts of a component.
 | `arch` | `string` | N |  | The CPU architecture required to host (all of) the component's containers (since containers share physical hardware with the underlying host). Possible values include:<ul><li>i386</li><li>amd64</li><li>arm</li><li>arm64</li></ul> Default can be none and let the runtime chose architecture. |
 | `containers` | [`[]Container`](#container) | N | | The OCI container(s) that implement the component. |
 | `workloadSettings` | [`Properties`](#properties) | N | | Declaration of non-container settings that should be passed to the workload runtime |
+| `expose` | [`[]Data`](#data) | N | | The data that the component will output |
```
hongchaodeng: _expose_ and _consume_ should be two different constructs.

Let me explain with a concrete example to start with:

```yaml
expose:
  name: abc
---
consume:
- name: abc
  as: env/file
```

There are two differences to note here:
1. _expose_ should output only one thing, while _consume_ could take many. The whole output of a component is exposed into one secret.
2. _expose_ doesn't  need _as_ field

wonderflow: I agree. I fixed them, and add global field in `Expose`

hongchaodeng:
lgtm

PR #294, opened 2020-01-03
fix the discrepancy in properties description in the spec

ryanzhang-oss:


hongchaodeng:
@resouer @wonderflow Could you also review it?

Issue #299, opened 2020-01-15
Missing specs for core traits


clareliguori:
The spec defines a single core trait `ManualScaler`, but seems to be missing additional core traits that are implemented by Rudr:
https://github.com/oam-dev/spec/blob/master/5.traits.md#core-traits

Rudr implements the following additional traits that are not defined in the spec:
- `ingress.core.oam.dev/v1alpha1`
- `auto-scaler.core.oam.dev/v1alpha1`
- `volume-mounter.core.oam.dev/v1alpha1`

The spec also states that the trait is named `ManualScaler`, but Rudr implements it as `manual-scaler`.

There seems to be a placeholder folder for these core trait specs, but there's nothing there yet:
https://github.com/oam-dev/spec/tree/master/standard/traits

hongchaodeng:
Hi @clareliguori 

Thanks for raising the question and I am happy to explain in details.

> The spec also states that the trait is named ManualScaler, but Rudr implements it as manual-scaler.

The spec is intended to be neutral and not bound to any specific platform. But when we implemented the spec on k8s the name cannot be CamelCase so we had to change it for convenience.

> additional traits that are not defined in the spec: ...

I agree with you that if we implemented something in Rudr, it'd better be defined in spec as well. Nonetheless, we are still working on defining what should be _core_. If we define something as _core_, that means all platforms need to implement them. We are in discretion because of this reason.

What traits should be _core_ in your mind? We would like to involve more community effort to define _core_ spec :)

resouer:
My feeling is maybe we can till define core traits as reference but don't have to enforce runtime must implement them all.

clareliguori:
> The spec is intended to be neutral and not bound to any specific platform. But when we implemented the spec on k8s the name cannot be CamelCase so we had to change it for convenience.

Are you open to changing the spec to remove the camel casing?  It would be great to be able to take the examples in the Rudr repo and use them with other spec-compliant implementations.

> Btw, what kind of traits could be core traits from your perspective? This input is super valuable.

Coming from the perspective of working on Amazon ECS, `ingress` feels widely applicable to many types of platforms and feels core to deploying many applications' Server components.  To me, that one should be a core trait.

`auto-scaler` can also be applied to ECS, and probably many other container platforms, so I support that in the core.  My only hesitation are platforms that don't auto-scale on metrics like CPU and memory (for example, platforms that auto-scale on requests), so perhaps only min and max should be required to be implemented.

[...]

PR #300, opened 2020-01-16
change trait name from CamelCase to snake_case

hongchaodeng:
ref: https://github.com/oam-dev/spec/issues/299

In Rudr the trait names are all in snake_case due to k8s naming restrictions.
To make the spec clear and consistent to users, we'd better fix all trait names to be snake_case.

### DIFF
--- a/5.traits.md
+++ b/5.traits.md
@@ -4,11 +4,11 @@ A _trait_ is a discretionary runtime overlay that augments a component workload 
 
 The _traits system_ is defined in this specification as the way a runtime applies and manages operational behaviors to instances of components through traits.
 
-Runtime implementations of the specification MUST provide the traits system for attaching operational behaviors to instances of components. 
+Runtime implementations of the specification MUST provide the traits system for attaching operational behaviors to instances of components.
 
-## Traits system rules and characteristics 
+## Traits system rules and characteristics
 
-Traits are applied to component instances in the [application configuration](6.application_configuration.md). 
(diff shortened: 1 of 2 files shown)


5.traits.md
```diff
@@ -153,14 +153,14 @@ spec:
       instanceName: web-front-end
       parameterValues:
       traits:
-        - name: ManualScaler
+        - name: manual-scaler.core.oam.dev/v1alpha1
```
resouer: This Trait name (i.e. this example from Rudr) does work but is essentially NOT right (i.e. Trait type is missing).

I'd suggest use right format here (e.g. update the spec and this example to include both Trait name and type), and send another PR to fix Rudr at the same time.

/cc @wonderflow

6.application_configuration.md
```diff
@@ -378,7 +378,7 @@ spec:
         - name: message
           value: "[fromVariable(message)]"
       traits:
-        - name: Ingress
+        - name: ingress.core.oam.dev/v1alpha1
```
resouer: same with all the other Trait names.

5.traits.md
wonderflow: yes

hongchaodeng:
Addressed comments

hongchaodeng:
Also fix the inconsistency in Rudr: https://github.com/oam-dev/rudr/pull/516

PR #304, opened 2020-02-06
OAM second working draft proposal

negz:
This design document iterates on the existing [Google Document on this topic](https://docs.google.com/document/d/1XBYj5Fg7A0gO6HYpvGqe2xPbm3iCulQfk9FMY6KCYV0/edit#heading=h.aerkwqfdauyp). Now that we're close to aligning on updates to the spec we'd like to move the discussion to GitHub.

### DIFF
--- a/design/20200105-spec-v1alpha2-kubernetes-friendly.md
+++ b/design/20200105-spec-v1alpha2-kubernetes-friendly.md
@@ -1,0 +1,483 @@
+# Kubernetes Friendly OAM
+
+* Author: Nic Cope ([@negz], [negz@upbound.io])
+* Reviewers: [OAM Maintainers]
+
+## Background
+
+The [Open Application Model] (OAM) is a specification for building cloud native
+applications. OAM aims to enable the [separation of concerns] between
+application developers, application operators, and infrastructure operators.
+[Crossplane] is an open source multicloud control plane that shares many of
+OAM's goals around enabling the separation of concerns when modelling cloud
+native applications.
+
+While OAM and Crossplane share goals, they take different approaches. Crossplane
+bets on the Kubernetes API server, leveraging it as its foundation. OAM is
+_inspired_ by the Kubernetes API, but intends to be implementable by runtimes
+that do not use Kubernetes at all. This runtime flexibility comes at a price to
+Kubernetes-native implementations; the OAM spec is suggestively _Kubernetes
+like_ but not particularly _idiomatic_ with regard to [Kubernetes API
+conventions]. For example:
+
+* Traits and Scopes, which define the schema for _parts of_ an ApplicationConfig
+  resource. In Kubernetes a resource’s schema is defined by its
+  [CustomResourceDefinition] (CRD).
+* ComponentSchematics, which can be either a template for configuration values
+  according to a fixed schema (the containers array), or define both schema and
+  configuration values (the workloadSettings object).
+
+Furthermore, the contemporary OAM specification cannot directly leverage the
+many custom resources that are already defined in the Kubernetes ecosystem;
+doing so requires that an OAM runtime understand how to parse an extended
+workload corresponding to each custom resource.
+
+Take, for example, a Crossplane _[resource claim]_:
+
+```yaml
+apiVersion: cache.crossplane.io/v1alpha1
+kind: RedisCluster
+metadata:
+  name: app-redis
+spec:
+  classSelector:
+    matchLabels:
+      region: "us-east"
+  writeConnectionSecretToRef:
+    name: app-redis-connection-details
+  engineVersion: "3.2"
+```
+
+The application of this RedisCluster resource claim to the API server causes
+Crossplane to provision a Redis cluster using an appropriate infrastructure
+provider. Resource claims are authored by application operators, like an OAM
+ApplicationConfig. They model the need for infrastructure from an application’s
+perspective; detailed configuration is defined by a _resource class_.
+
+Ideally resource claims (and other Kubernetes custom resources like them) could
+be first class citizens of OAM. Conversely, the contemporary OAM specification
+imposes a layer of abstraction; an extended workload type. Crossplane (as a
+tentative OAM runtime) would need to define an extended workload type
+corresponding to the above resource claim:
+
+```yaml
+apiVersion: core.oam.dev/v1alpha1
+kind: ComponentSchematic
+metadata:
+  name: rediscluster
+spec:
+  workloadType: cache.crossplane.io/v1alpha1.RedisCluster
+  parameters:
+  - name: region
+    type: string
+    required: true
+  - name: engineVersion
+    type: string
+  workloadSettings:
+    - name: engineVersion
+      type: string
+      fromParam: region
+    - name: region
+      type: string
+      fromParam: region
+```
+
+In addition to defining this extended workload type, the OAM runtime would need
+to know how to render it, i.e. how to produce a RedisCluster custom resource
+given a ComponentSchematic of the RedisCluster workload type. This rendering
+process would presumably require the OAM runtime to contain logic for each
+custom resource.
+
+## Goals
+
+OAM as a specification is heavily inspired by the Kubernetes API, but is runtime
+agnostic. The purpose of this document is to explore whether the OAM
+specification could be made friendlier to - idiomatic to and compatible with -
+Kubernetes powered runtimes while remaining flexible enough to be usefully
+implemented by other runtimes.
+
+Kubernetes powered runtimes are the focus of this document because the OAM
+specification already looks and feels a lot like the Kubernetes API. When it is
+implemented using the _actual_ Kubernetes API any divergences from said API’s
+conventions are an order of magnitude more likely to violate the [principle of
+least astonishment] for users than for users of other runtimes.
+
+Any proposal outlined by this document should ensure:
+
+* The separation of concerns that is the core focus of the OAM specification.
+* That the OAM specification “look and feel” idiomatic to the Kubernetes API
+  model.
+* That the OAM specification remains useful and compelling to other runtimes.
+
+Note that this document frequently uses Crossplane as an example of a tentative
+OAM runtime that would benefit from a Kubernetes friendly OAM specification, but
+defining specifically how Crossplane will support OAM is out of scope.
+
+## Proposal
+
+This document proposes:
+
+1. An ApplicationConfiguration _references_ distinct Component resources,
+   grouping them into one logical application and providing inputs to their
+   parameters.
+1. Each Component embed an OAM “workload” under the `.spec.workload` key.
+   Available workloads are defined by WorkloadDefinitions. In Kubernetes
+   runtimes any existing custom resource with a corresponding WorkloadDefinition
+   can be used in a Component.
+1. The set of available Traits and their schemas be defined via
+   TraitDefinitions. Traits are applied to a component and configured by
+   specifying their configuration inline in the ApplicationConfig as an
+   "embedded resource".
+1. The set of available Scopes and their schemas be defined via
+   ScopeDefinitions. Scopes are referenced by an ApplicationConfiguration.
+
+### Components and Workloads
+
+Each Component is of a distinct workload _kind_. This is functionally equivalent
+to the existing concept of workload _types_, but reuses the _kind_ term already
+used within Kubernetes and OAM. The schema of a workload kind - including core
+workload kinds like Server - are defined via a WorkloadDefinition. For example:
+
+```yaml
+apiVersion: core.oam.dev/v1alpha1
+kind: WorkloadDefinition
+metadata:
+  name: server.core.oam.dev
+spec:
+  definitionRef:
+    name: server.core.oam.dev
+```
+
+The above WorkloadDefinition references a Kubernetes CustomResourceDefinition
+named server.core.oam.dev, designating custom resources defined by said CRD to
+be valid OAM workloads. OAM runtimes that are not powered by Kubernetes could
+opt to either implement the CRD concept, or introduce a new field to the
+WorkloadDefinition spec that defines a kind of workload and its schema.
+Application operators can discover available OAM workload kinds by running
+`kubectl get workloaddefinitions`. Workloads would most frequently be defined by
+OAM runtimes or infrastructure operators.
+
+```yaml
+apiVersion: core.oam.dev/v1alpha1
+kind: Component
+metadata:
+  name: example-server
+spec:
+  workload:
+    apiVersion: core.oam.dev/v1alpha1
+    kind: Server
+    spec:
+      osType: linux
+      containers:
+      - name: my-cool-server
+        image:
+          name: example/very-cool-server:1.0.0
+          digest: sha256:verytrustworthyhash
+        resources:
+          cpu:
+            required: 1.0
+          memory:
+            required: 100MB
+        ports:
+        - name: http
+          value: 8080
+        env:
+        - name: CACHE_SECRET
+        livenessProbe:
+          httpGet:
+            port: 8080
+            path: /healthz
+        readinessProbe:
+          httpGet:
+            port: 8080
+            path: /healthz
+  parameters:
+  - name: instanceName
+    required: true
+    fieldPaths:
+    - ".metadata.name"
+  - name: cacheSecret
+    required: true
+    fieldPaths:
+    - ".spec.containers[0].env[0].value"
+```
+
+Components would most frequently be defined by application developers. A
+Component has two responsibilities:
+
+1. Embed a valid OAM workload kind under its `.spec.workload` field.
+2. Publish a set of parameters under its `.spec.parameters` field. Parameters
+   may be set by an ApplicationConfig. Their values overlay onto fields in the
+   embedded OAM workload.
+
+The embedded OAM workload is a template of a sorts - though it is important to
+note that it is a first class citizen of the configuration language. It is
+specified using YAML (or JSON, etc), not as an opaque string. OAM runtimes can
+validate its schema against its WorkloadDefinition. Kubernetes runtimes may
+consider the workload an [embedded object], like a polymorphic podTemplateSpec.
+
+OAM runtimes would use the set of parameters published by a Component to mutate
+the embedded workload based on inputs supplied by the ApplicationConfiguration.
+Note that the relationship between parameters and workload fields is inverted
+compared to the contemporary OAM spec. Each parameter includes one or more JSON
+field paths specifying that its value should be injected at those paths. This
+[existing] [Kubernetes paradigm] ensures the parameters supported by a component
+are easily discoverable, and allows the OAM runtime to inject parameters into
+any field of the embedded workload without introducing a [“fromParams()” inline
+function DSL]. This approach also allows the published parameters to omit their
+type - the type can be inferred from the type of the workload fields a parameter
+corresponds to.
+
+```yaml
+apiVersion: core.oam.dev/v1alpha1
+kind: WorkloadDefinition
+metadata:
+  name: rediscluster.cache.crossplane.io
+spec:
+  definitionRef:
+    name: rediscluster.cache.crossplane.io
+```
+
+```yaml
+apiVersion: core.oam.dev/v1alpha1
+kind: Component
+metadata:
+  name: example-cache
+spec:
+  workload:
+    apiVersion: cache.crossplane.io/v1alpha1
+    kind: RedisCluster
+    spec:
+      engineVersion: "3.2"
+  parameters:
+  - name: instanceName
+    required: true
+    fieldPaths:
+    - ".metadata.name"
+  - name: secret
+    required: true
+    fieldPaths:
+    - ".spec.writeConnectionSecretToRef.name"
+  - name: engineVersion
+    required: false
+    fieldPaths:
+    - ".spec.engineVersion"
+```
+
+The above WorkloadDefinition and Component leverages the existing Crossplane
+RedisCluster resource claim. Crossplane as a tentative OAM runtime need not
+define an extended workload type; it need only create a WorkloadDefinition
+pointing to [the existing CustomResourceDefinition] of the RedisCluster resource
+claim.
+
+The Component design enables OAM runtimes to break the task of reconciling the
+desired state of an application with the actual state of the world into two
+distinct stages:
+
+1. Render workloads. The OAM runtime iterates over the Components referenced by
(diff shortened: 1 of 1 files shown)


resouer:
@negz Thanks for quick PR. Awesome.

Can we give this proposal a formal name like "OAM spec proposal 0.3.0" or sth? /cc @vturecek

design/kubernetes-friendly-oam.md
```diff
@@ -0,0 +1,478 @@
...
+```yaml
+apiVersion: core.oam.dev/v1alpha1
+kind: ApplicationConfiguration
+metadata:
+  name: cool-example
+spec:
+  components:
+  - componentName: example-server
+    traits:
+    - apiVersion: core.oam.dev/v1alpha1
```
hongchaodeng: We should put a trait CR under an envelope field:

```
traits:
- name: manual-scaler
  instance:
    ...
```

In this way, we could achieve extensibility and add other models (ref/properties) later.

design/kubernetes-friendly-oam.md
```diff
@@ -0,0 +1,478 @@
...
+
+Note that this document frequently uses Crossplane as an example of a tentative
+OAM runtime that would benefit from a Kubernetes friendly OAM specification, but
+defining specifically how Crossplane will support OAM is out of scope.
+
+## Proposal
+
+This document proposes:
+
+1. An ApplicationConfiguration _references_ distinct Components and Trait
```
xiang90: is this still true? in this example https://github.com/oam-dev/spec/pull/304/files#diff-01be24953bc11e69526ef7b88a72f482R420, only scope uses the reference model.

negz: Good catch - it's no longer true. I'll update it.

design/kubernetes-friendly-oam.md
negz: I like the idea of using an envelope field, though I'd prefer to avoid duplicating the trait's name (we could already use its object metadata for that). I'll make an update.

design/kubernetes-friendly-oam.md
```diff
@@ -0,0 +1,478 @@
...
+```yaml
+apiVersion: core.oam.dev/v1alpha1
+kind: ApplicationConfiguration
+metadata:
+  name: cool-example
+spec:
+  components:
+  - componentName: example-server
+    parameterValues:
+    - name: instanceName
```
hongchaodeng: The `name-value` pair is unnatural to yaml users. It is just unfit here and looks like XML.

Could we change them to a map instead:
```yaml
parameterValues:
  instanceName: cool-example
  cacheSecret: cache-connection
```

xiang90: then we need to enforce that no `:` in the parameter name.

vturecek:
@resouer done.

design/kubernetes-friendly-oam.md
vturecek: you also can't schematize it if you change it to a map, since each component would have different parameters.

design/kubernetes-friendly-oam.md
vturecek: Wouldn't it make more sense if the trait was referring to a TraitDefinition? I'm imagining a workflow where an app operator asks the runtime for all available trait definitions, and then adds traits in the app config by picking from that list. In that case, they'd probably be looking at the name metadata field in the TraitDefinition. The runtime would also need to know the TraitDefinition to check if the trait supports the workload type of the component it's being applied to.

design/kubernetes-friendly-oam.md
```diff
@@ -0,0 +1,478 @@
...
+spec:
+  definitionRef:
+    name: server.core.oam.dev
+```
+
+The above WorkloadDefinition references a Kubernetes CustomResourceDefinition
+named server.core.oam.dev, designating custom resources defined by said CRD to
+be valid OAM workloads. OAM runtimes that are not powered by Kubernetes could
+opt to either implement the CRD concept, or introduce a new field to the
+WorkloadDefinition spec that defines a kind of workload and its schema.
```
abchow: Hi Nic, Nice proposal!  I can see there are a lot of thoughts in it.   The unique features of K8N are leveraged and I can see it married well with OAM schema.  
	
I am kind of confused in this section.  Hope that you can help.
	
I believe at one point, we are saying the "Core" schema will be enforced; while the "Extended" will be open up for different providers.  This seen to be the Core Server workload definition, but what does it means by "introduce a new field to the WorkloadDefinition spec that defines a kind of workload and its schema"?  Does it means Developers or Application Operators need to specify parameter based on the defined schema?

design/kubernetes-friendly-oam.md
resouer: @vturecek I think Nic's current example also make sense. e.g. if the TraitDefinition is referring to a CRD, then the Operator could embed a CR in this section (inside the envelope).

design/kubernetes-friendly-oam.md
negz: Using a map would also [violate API conventions](https://github.com/kubernetes/community/blob/7e1b70/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps) for Kubernetes runtimes. I'd prefer to stick with the list of named subobjects.

design/kubernetes-friendly-oam.md
negz: > I like the idea of using an envelope field

I like this idea in principle, but it feels awkward in practice. See the below examples where I've tried embedding the CR under the key `trait` and `instance`.

```yaml
metadata:
  name: cool-example
spec:
  components:
  - componentName: example-server
    traits:
      trait:
      - apiVersion: core.oam.dev/v1alpha1
        kind: ManualScalerTrait
        spec:
          replicaCount: 3
```

```yaml
metadata:
  name: cool-example
spec:
  components:
  - componentName: example-server
    traits:
      instance:
      - apiVersion: core.oam.dev/v1alpha1
        kind: ManualScalerTrait
        spec:
          replicaCount: 3
```

> In this way, we could achieve extensibility and add other models (ref/properties) later.

I think this is really the only case in which it makes sense to put the trait CR under an envelope field. In any other case we have an extensibility story by adding fields to the trait CRs themselves. How confident are we that we'll want to go down this path some day? I still prefer to avoid it, because I think the spec will be simpler if there's only one way to do things (like use traits).

negz: > Wouldn't it make more sense if the trait was referring to a TraitDefinition?

It's not immediately obvious to me that it would make more sense - I think the UX of authoring ApplicationConfiguration is better when we're embedding a trait (not referencing a TraitDefinition), but I can see your point about trait discoverability. One thing we could do to help with this would be to use [custom printer columns](https://book.kubebuilder.io/reference/markers.html) for our `TraitDefinition` CRD such that folks running kubectl would see something like:


```
$ kubectl get traitdefinitions
NAME                  KIND               API-VERSION
manualscaler.oam.dev  ManualScalerTrait  core.oam.dev/v1alpha1
```

design/kubernetes-friendly-oam.md
negz: > what does it means by "introduce a new field to the WorkloadDefinition spec that defines a kind of workload and its schema"? Does it means Developers or Application Operators need to specify parameter based on the defined schema?

The goal here was to point out that OAM runtimes that are not powered by Kubernetes could choose to add more settings related to workloads by adding new optional fields to the spec of a `WorkloadDefinition` (for example a field that defined an inline schema instead of pointing to a `definitionRef`). I don't expect application developers or operators will need to interact with workload definitions in any way apart from browsing them to discover which workloads are available and what their parameters are. Does that make sense?

abchow: Hi Nic, thanks for the clarification.
From how I understand, Component become instance of Workload.  
If we add settings to a workload, we are expecting developer or operator to set parameters to those settings.
In this case, the settings schema of Core Workloads become a concern.  
In the WorkloadDefinition example above, "server.core.oam.dev" is a CRD and it is definitionRef.  
Am I correct that the settings of the parameters, for the Component example below, comes from "server.core.oam.dev"?
Also, "a field that defined an inline schema instead of pointing to a definitionRef".  
These sound like we are allowing workload implementations to define their own setting schema.
Is this correct?

design/kubernetes-friendly-oam.md
hongchaodeng: Why not embed like this?

```yaml
metadata:
  name: cool-example
spec:
  components:
  - componentName: example-server
    traits:
      - name: manual-scaler
        instance:
          kind: ManualScalerTrait
          spec:
            replicaCount: 3
```

design/kubernetes-friendly-oam.md
negz: We discussed and clarified this in this morning's community meeting. For posterity, I believe @abchow's primary concern was that core workload kinds would have identical schemas across OAM runtimes; i.e. you would configure a `server.oam.dev/v1alpha` workload kind exactly the same on a Kubernetes OAM runtime as you would on an AcmeCloud OAM runtime. I believe this would be the case.

design/kubernetes-friendly-oam.md
```diff
@@ -0,0 +1,478 @@
...
+`kubectl get workloaddefinitions`. Workloads would most frequently be defined by
+OAM runtimes or infrastructure operators.
+
+```yaml
+apiVersion: core.oam.dev/v1alpha1
+kind: Component
+metadata:
+  name: example-server
+spec:
+  workload:
```
negz: @vturecek @hongchaodeng @ryanzhang-oss @resouer - This morning we spoke in the community meeting about how application operators would specify traits in the OAM ApplicationConfiguration. I believe we decided that the inline traits would need to be somehow relatable back to their trait definitions (e.g. by including some kind of type metadata), but would not explicitly need to include `kind` and `apiVersion` metadata.

How does this relate to workload kinds, especially core ones such as `Server`? We're proposing here that the `Component` embed a core workload type, which looks like a Kubernetes CR. The workload's kind is defined by the `apiVersion` and `kind` fields of the embedded CR. My thinking was that these core workload kinds would look like a Kubernetes CR (with `apiVersion` and `kind) even if the OAM runtime were not Kubernetes based. This allows portability of `Component` instances across runtimes - which I think was @abchow's concern this morning.

design/kubernetes-friendly-oam.md
negz: > Why not embed like this?

We discussed this in the OAM community meeting this morning, but for posterity:

* I think `name` is redundant, whether it's meant to be the name of the trait definition or the name of the trait instance that will be created. The former isn't needed at the ApplicationConfiguration level - the AppConfig controller will just instantiate the templated trait CR, whose separate controller will check its TraitDefinition. The latter duplicates the trait's existing `.metadata.name` field.
* I feel like the name `instance` for an envelope key "leaks implementation details" of the spec debate process, or put differently, doesn't make much sense to those of us who haven't worked on the spec.

We agreed that we do want to use an envelope field just in case, and that we'd name that field `trait`. I'll make the updat.e

design/kubernetes-friendly-oam.md
abchow: Thanks Nic.  For clarification, my concern was the Settings / Parameters of Core types are not enforced.  If it is not enforced, different OAM runtime providers can provide their own setting attributes.  For example, Provider A can call container image as "imageLoc"; while Provider B can call it "swrImage".  In this case, the model that work in Provider A will not work on Provider B without modification, and it defeated the goal of Platform-Independent.
It was make clear to me that this proposal will not affect the Settings / Parameters list of the Core types.  Thanks

design/kubernetes-friendly-oam.md
resouer: The indexing mechanism from `workload -> def'n` and `trait -> def'n` is same IMO.

hongchaodeng:
LGTM

It seems that we have agreed on the spec. We can merge this and fix the minor issues later.

design/kubernetes-friendly-oam.md
negz: @resouer It's not clear to me what you mean.

My point here is that we did not seem comfortable with explicitly requiring traits to include an `apiVersion` and `kind` in their schema, yet we seem to be doing that implicitly for workloads. A workload on Kubernetes is an embedded CR, and CRs must have `apiVersion` and `kind`. We want to have a set of core workloads (like the `kind: Server` example here) that are portable across runtimes. It therefore follows that at least our core workloads _must_ have `apiVersion` and `kind` in their schema regardless of what platform they're on. Why not take this one step further and require that all traits, scopes, and workloads contain `apiVersion` and `kind` regardless of runtime?

resouer: It's not the intention to make them different. So what I mean is as long as we agree on one approach, it's reasonable applies to both workload and trait (and other xxxDefinition enforced concepts)

design/kubernetes-friendly-oam.md
```diff
@@ -0,0 +1,482 @@
...
+it should apply to. For example:
+
+```yaml
+apiVersion: core.oam.dev/v1alpha1
+kind: ApplicationConfiguration
+metadata:
+  name: cool-example
+spec:
+  components:
+  - componentName: example-server
```
wonderflow: why we don't need `instanceName` here?

zhxu2: as I was reading the google doc, instance name seems to have been put under parameterValues

design/kubernetes-friendly-oam.md
```diff
@@ -0,0 +1,482 @@
...
+    - name: instanceName
+      value: cool-example
+    - name: cacheSecret
+      value: cache-connection
+    traits:
+      trait:
+      - apiVersion: core.oam.dev/v1alpha1
+        kind: ManualScalerTrait
+        spec:
+          replicaCount: 3
```
wonderflow: I think the format here should change like this:

```
    traits:
      - trait:
           kind: ManualScalerTrait
           spec:
             replicaCount: 3
```

This is like scopeRef

design/kubernetes-friendly-oam.md
```diff
@@ -0,0 +1,482 @@
...
+metadata:
+  name: manualscalertrait.core.oam.dev
+spec:
+  appliesToWorkloads:
+    - server.core.oam.dev/v1alpha1
+    - worker.core.oam.dev/v1alpha1
+    - task.core.oam.dev/v1alpha1
+  definitionRef:
+    name: manualscalertrait.core.oam.dev
+```
```
wonderflow: besides definitionRef, we'd better add `definition` field here. Then we could allow non-K8s implementation to define json schema here.

wonderflow:
``` yaml
kind: TraitDefinition
metadata:
  name: manualscalertrait.core.oam.dev
spec:
  appliesToWorkloads:
    - server.core.oam.dev/v1alpha1
    - worker.core.oam.dev/v1alpha1
    - task.core.oam.dev/v1alpha1
  definitionRef:
    name: manualscalertrait.core.oam.dev
  definition: |
      {
      "$schema": "http://json-schema.org/draft-07/schema#",
      "type": "object",
      "required": [
        "replicaCount"
      ],
      "properties": {
        "replicaCount": {
          "type": "integer",
          "description": "the target number of replicas to scale a component to.",
          "minimum": 0
        }
      }
    }
```

design/kubernetes-friendly-oam.md
```diff
@@ -0,0 +1,482 @@
...
+kind: TraitDefinition
+metadata:
+  name: manualscalertrait.core.oam.dev
+spec:
+  appliesToWorkloads:
+    - server.core.oam.dev/v1alpha1
+    - worker.core.oam.dev/v1alpha1
+    - task.core.oam.dev/v1alpha1
+  definitionRef:
+    name: manualscalertrait.core.oam.dev
```
wonderflow:
```suggestion
    indexName: v1alpha1.manualscalertrait.core.oam.dev
```

design/kubernetes-friendly-oam.md
```diff
@@ -0,0 +1,482 @@
...
+particular kind:
+
+```yaml
+apiVersion: core.oam.dev/v1alpha1
+kind: ScopeDefinition
+metadata:
+  name: networkscope.core.oam.dev
+spec:
+  allowComponentOverlap: true
+  crdName: networkscope.core.oam.dev
```
wonderflow: we'd better align with trait:

```
kind: ScopeDefinition
metadata:
  name: networkscope.core.oam.dev
spec:
  allowComponentOverlap: true
  definitionRef:
     indexName: networkscope.core.oam.dev
```

and for non-K8s:
kind: ScopeDefinition
metadata:
  name: networkscope.core.oam.dev
spec:
  allowComponentOverlap: true
  definition: |
     {
      "$schema": "http://json-schema.org/draft-07/schema#",
      "type": "object",
      "required": [
        "replicaCount"
      ],
      "properties": {
        "networkId": {
          "type": "string",
          "description": "subnet id of this network scope"
        }
      }
    }
```

```

design/kubernetes-friendly-oam.md
wonderflow: Or
```
name: manualscalertrait.core.oam.dev
version: v1alpha1
```

design/kubernetes-friendly-oam.md
wonderflow: after discuss with @ryanzhang-oss , we both agree that non-k8s implementation can also use definitionRef. So we'd better not add `definition`.

design/kubernetes-friendly-oam.md
wonderflow: After deep consideration, I think we'd better use `name` and `version` in definitionRef.

```
    traits:
      - name: manualscalertrait.core.oam.dev
         version: v1alpha1
         schematic:
           replicaCount: 3
```

I have these reasons:

1. Same experience with traitDefinition. Users could easily learn how to write, not bother learn schema twice.
2. In the OAM Spec, we just need to leave `schematic` to be custom defined. This could not only be more clean, but also be easier for non-k8s implementation to define the schematic in traitDefinition.
3. The way we validate the trait in application configuration will be more intuitive for both K8s and non-K8s implementation. We follow the schema from trait definition (in fact it's definitionRef), then we can index the schema directly from definitionRef with name and version. If you use apiVersion and kind here, just think for non-k8s implementation, they have to implement same mechanism like this.

wonderflow:
Sorry for commenting so much after a `LGTM`. As I was changing the spec according this PR, so I find something still need to change. Hope we could resolve them ASAP.

design/kubernetes-friendly-oam.md
resouer: This is not align with what we discussed in the meeting. We agreed to use K8s resource model for end user facing objects (specially, for trait spec, we will recommend this model). The intention is to eliminate confusion and create a "k8s-like" resource here which is not friendly to either K8s or non-K8s implementation.

design/kubernetes-friendly-oam.md
resouer: I'd suggest we do not distribute discussion into multiple issues. 

Also, this LGTM because `version` could be an optional field.

design/kubernetes-friendly-oam.md
resouer: +1

resouer:
I think the only change this proposal need to be updated is this inconsistency: https://github.com/oam-dev/spec/pull/304#discussion_r383411794

Generally LGTM, could be merged as long as above nits been fixed. Thanks Nic.

design/kubernetes-friendly-oam.md
negz: > I think the format here should change like this

Good call! What you propose in your first comment is what I had intended to write; i.e. the `traits` field should contain an array. I'll update.

design/kubernetes-friendly-oam.md
negz: I'd like to ensure this remains a strict subset of the `ObjectReference` type in accordance with API conventions per https://github.com/kubernetes/community/blob/1d6d12/contributors/devel/sig-architecture/api-conventions.md#references-to-related-objects.

design/kubernetes-friendly-oam.md
negz: Yeah, this should have been changed to `definitionRef` too - will fix.

design/kubernetes-friendly-oam.md
ryanzhang-oss: I wonder if there is any particular reason to adhere to `ObjectReference` type?  Is the controller translate our definitionRef to it?

negz: @ryanzhang-oss It's an API consistency constraint, not a technical one. So the reason is mostly just that Kubernetes API conventions bless this pattern as the standard one, and thus it's the least surprising option for folks interacting with our very-Kubernetes-like API.

wonderflow:
Cool, I think we could merge it now.

design/kubernetes-friendly-oam.md
```diff
@@ -0,0 +1,483 @@
...
+    - name: cacheSecret
+      value: cache-connection
+    traits:
+    - trait:
+        apiVersion: core.oam.dev/v1alpha1
+        kind: ManualScalerTrait
+        spec:
+          replicaCount: 3
+    scopes:
+    - scopeRef:
```
resouer: or just `scope`?

wonderflow: yeah, align with trait

negz: I think this should stay as `scopeRef`. It's a _reference to a scope_, whereas a trait is _an embedded trait_.  Per API conventions (I feel like a broken record referring to these conventions) fields that are a reference to a `kind: Foo` should be named `fooRef`.

https://github.com/kubernetes/community/blob/96203/contributors/devel/sig-architecture/api-conventions.md#naming-conventions

resouer:
LGTM, we are ready to ship this proposal except one requirement:

@negz Could you please give this file to a formal name? I'd expect we have more proposals in the future so let's make them organized.

My idea is: 20200120-1.0.0-alpha2-oam-spec.md (borrowed from KEP template: YYYYMMDD-title.md)

negz:
>Could you please give this file to a formal name?

@resouer  I like the idea of using the KEP format and mentioning that the result is a new v1alpha2 version of the spec. I also think it's important to capture the context around making the spec "Kubernetes friendly", since that was the driver behind these changes. I went with `20200105-spec-v1alpha2-kubernetes-friendly.md` - does that sound okay?

resouer:
LGTM

PR #311, opened 2020-02-11
add Ryan to the owner list

ryanzhang-oss:


resouer:
Ryan has contributed a lot to the new version of spec as well as K8s implementation (Rudr).

LGTM

PR #312, opened 2020-02-12
Adjust text according to new spec

ryanzhang-oss:
Move workload type up in the list

### DIFF
--- a/1.purpose_and_goals.md
+++ b/1.purpose_and_goals.md
@@ -1,19 +1,19 @@
 # 1. Purpose and Goals
 The Open Application Model defines a standard, platform-agnostic way to describe cloud and edge applications. The goal of the specification is to provide a common way to describe applications agnostic to any specific container runtime, orchestration software, cloud provider, or hardware configuration, with clearly defined roles for developers and operators, while still allowing implementations to make use of the native APIs, tools, and features that are unique to the implementation and its underlying platform.  
 
-The central problem this specification seeks to address is how distributed applications can be composed and then successfully handed off to those responsible for operating them. The problem is not so much about how to write programs, but how to take components of a service oriented (or microservice oriented) architecture, and streamline the workflow that surrounds such applications.
+The central problem this specification seeks to address is how distributed applications can be composed and handed off to those responsible for operating them. The problem is not so much about how to write programs, but how to take components of a service oriented (or microservice oriented) architecture, and streamline the workflow that surrounds such applications.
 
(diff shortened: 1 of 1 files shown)


1.purpose_and_goals.md
```diff
@@ -1,19 +1,19 @@
...
+This specification provides a description of such applications, where the description is aimed at separating the concerns of developers from the concerns of operators. Furthermore, it suggests patterns and processes for managing such applications. The specification describes a model for cloud native (i.e. highly distributed) applications, encompassing public cloud technologies, on-prem solutions, and IoT/edge technologies. By specifying a common model, this specification provides a foundation for multiple implementations of a shared description of cloud native applications.
 
 This open specification defines:
 
-- __Component schematics__, where developers declare the operational characteristics of the code they deliver _in infrastructure neutral terms_.
+- __Workload types and definitions__, where platforms declare the schema of a specific type of workload
```
negz: How would you feel about using the term "workload kind" instead of "workload type"? They're semantically equivalent, but "kind" fits better with the fact that workloads will now be `kind: Server`, not `type: Server` in configuration documents.

1.purpose_and_goals.md
```diff
@@ -22,13 +22,13 @@ Additional goals of the specification are to:
 
 Non-goals include:
 
-- Defining or prescribing specific development or operational workflows.
+- Defining or prescribing specific development or operational workflow.
```
negz: I believe this was more grammatically correct with the "s". "Defining or prescribing _a_ specific development or operational workflow" would also read well.

1.purpose_and_goals.md
```diff
@@ -22,13 +22,13 @@ Additional goals of the specification are to:
...
 - Defining the schemas of operational resources, for example (but not limited
   to):
   - Secrets (secure, encrypted values)
   - Networks
   - Volumes
-- Describing or defining the runtimes themselves
+- Describing or defining the runtime themselves
```
negz: I think this was more grammatically correct with the "s" too, but "Describing or defining a runtime itself" would also work.

1.purpose_and_goals.md
negz: https://github.com/oam-dev/spec/pull/304/files#diff-01be24953bc11e69526ef7b88a72f482R136

More context from the draft proposal here.

ryanzhang-oss: thx,  I tweaked the wording a bit more, please take a look

hongchaodeng:
Since it is alpha2 work and minor change, merging now.

PR #316, opened 2020-02-22
[1.0.0-alpha2] Workload and component

ryanzhang-oss:
Revamped all the workload and component part.


4.component.md
```diff
@@ -491,156 +70,49 @@ metadata:
...
-      httpGet:
-        port: 8080
-        path: /healthz
-    readinessProbe:
-      httpGet:
-        port: 8080
-        path: /healthz
-```
-
-#### Example: Extended Workload Type
```
wonderflow: we also need to add an extended workload and component here

ryanzhang-oss: We don't have a good extended workload definition (schema) and I don't want to add a CRD example in our spec

2.overview_and_terminology.md
```diff
@@ -64,7 +64,7 @@ A trait is described as:
 
 - Metadata: Information about the trait
 - Applies-to list: Enumeration of which workload this trait applies to
-- Properties: Configuration that may be specified in an application configuration
+- Schema reference: A reference to a schema that defines the trait
```
resouer: I'm in favor of:

```
Definition reference: A reference to a schema that defines the trait.
```
The intention is to align with DefinitionRef.

2.overview_and_terminology.md
```diff
@@ -73,8 +73,8 @@ Application scopes provide different ways to group components into applications.
 An application scope is described as:
 
 - Metadata: Information about the scope
-- Type: The type of scope. Different types of scopes provide different capabilities.
-- Parameters: Configuration that may be specified in an application configuration
+- Allow overlap: Specifies whether a component can exist in multiple type of this scope
```
resouer: Maybe we need to keep the Type of scope? 

Because in OAM impl, `allowComponentOverlap` should only check if given Component is included by Scopes of same type only.

2.overview_and_terminology.md
```diff
@@ -281,5 +281,5 @@ This form is not accepted as an alternative for the fully qualified version. It
 
 | Previous Part        | Next Part          |
 | ------------- |-------------|
-|[1. Purpose And Goals](1.purpose_and_goals.md)| [3. The Component Model](4.component.md)|
+|[1. Purpose And Goals](1.purpose_and_goals.md)| [3. Worklaod Definition](3.workload.md)|
```
resouer: It seems we deleted Component model doc but added back WorkloadDefinition only? Are we going to leave Component for next PR?

4.component.md
```diff
@@ -74,415 +47,21 @@ The Parameters section defines all of the configurable parameters for this compo
...
-
-Because this component is stateless, horizontally scalable, and exposes an endpoint, it is best described
-as a `Server`. The username, password, and address of the backend
-component are configurable.
+This section illustrates use of the component schematic defined above to describe a component.
 
 ```yaml
-apiVersion: core.oam.dev/v1alpha1
+apiVersion: core.oam.dev/v1alpha2
 kind: ComponentSchematic
```
wonderflow:
```suggestion
kind: Component
```

3.workload.md
```diff
@@ -0,0 +1,378 @@
+# 3. Workload
+
+This section defines the workload definition schematics.
+
+Workload definitions are provided by the platform (just as traits are) so that users may inspect the platform and
+ learn what workloads are available for use. Workload definitions are not extensible to platform users (only to platform
+  operators). Thus, platform users MUST NOT be allowed to create new workload kinds.
```
resouer: We need to use terms like `Application Developers` etc in the whole doc instead of "platform users" etc for better consistency.

Another suggestion to make above narrative clearer:

Workload Definition is provided by the platform to enforce and validate the workload specification of given `Component`. It's designed to be a platform level concept and expected to be managed by Infrastructure Operators only. The OAM platform should provide a proper approach (e.g. docs or samples) to help Application Developers define `spec.workload` section in Component instead of exposing Workload Definitions directly to them.

3.workload.md
```diff
@@ -0,0 +1,378 @@
+# 3. Workload
+
```
resouer: Maybe we can add a `**Target User: Infrastructure Operator**` section here (and to the beginning of every other doc)?

3.workload.md
```diff
@@ -0,0 +1,378 @@
...
+This section defines the workload definition schematics.
+
+Workload definitions are provided by the platform (just as traits are) so that users may inspect the platform and
+ learn what workloads are available for use. Workload definitions are not extensible to platform users (only to platform
+  operators). Thus, platform users MUST NOT be allowed to create new workload kinds.
+
+## Top-Level Attributes
+
+The following attributes are common across all schemata defined in this document. This structure is designed to allow
```
resouer: I hope we can delete this section from every doc, and link/copy Kubernetes resource model documentation here (https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/#required-fields).

3.workload.md
```diff
@@ -0,0 +1,378 @@
...
+| Attribute | Type | Required | Default Value | Description |
+|-----------|------|----------|---------------|-------------|
+| `apiVersion` | `string` | Y | | The specific version of the specification in use. The core types use `core.oam.dev/v1alpha2` |
+| `kind` | `string` | Y || The string `WorkloadDefinition` |
+| `metadata` | [`Metadata`](#metadata) | Y | | WorkloadDefinition metadata. |
+| `spec`| [`Spec`](#spec) | Y | | A container for the workload definition. |
+
+### Metadata
+
+The metadata describes this workload definition. The `metadata` section is defined in [the meta section of part 2](2.overview_and_terminology.md#Metadata)
```
resouer: I also hope we can directly reference `metadata` from Kubernetes resource model (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#objectmeta-v1-meta) in "overview and terminology" doc. 

I'm not in favor of making the spec tedious and complex due to explanations on resource model and metadata.

3.workload.md
```diff
@@ -0,0 +1,378 @@
...
+| `DefinitionRef` | [`DefinitionRef`](#DefinitionRef) | Y | | Workload schema reference. |
+
+
+#### DefinitionRef
+
+DefinitionRef defines where to find the workload schema
+
+| Attribute | Type | Required | Default Value | Description |
+|-----------|------|----------|---------------|-------------|
+| `name` | `string` | Y | | The workload schema's name. Must be formatted in a way that can uniquely identify a specific version of the definition  |
```
resouer: We really need sample to make spec easy to understand.

Also I'm not sure that "indexing" works for non-CRD schema. I'm thinking maybe:

```yaml
definitionRef:
  name: scaling.alibaba.com/v1alpha2.Manualscaling
  schematic: "json schema string" # this is a optional field only needed for non-CRD platforms
```

3.workload.md
```diff
@@ -0,0 +1,378 @@
...
+| `metadata` | [`Metadata`](#metadata) | Y | | WorkloadDefinition metadata. |
+| `spec`| [`Spec`](#spec) | Y | | A container for the workload definition. |
+
+### Metadata
+
+The metadata describes this workload definition. The `metadata` section is defined in [the meta section of part 2](2.overview_and_terminology.md#Metadata)
+
+### Spec
+
+The specification section contains a reference to the workload schema
```
resouer:
```suggestion
The specification section contains a reference of the workload's definition.
```

3.workload.md
```diff
@@ -0,0 +1,378 @@
...
+
+| Attribute | Type | Required | Default Value | Description |
+|-----------|------|----------|---------------|-------------|
+| `DefinitionRef` | [`DefinitionRef`](#DefinitionRef) | Y | | Workload schema reference. |
+
+
+#### DefinitionRef
+
+DefinitionRef defines where to find the workload schema
```
resouer:
```suggestion
DefinitionRef defines the schema of workload. Note that it could be simply an indexing for a schema.
```

3.workload.md
```diff
@@ -0,0 +1,378 @@
...
+| Attribute | Type | Required | Default Value | Description |
+|-----------|------|----------|---------------|-------------|
+| `name` | `string` | Y | | The workload schema's name. Must be formatted in a way that can uniquely identify a specific version of the definition  |
+
+The workload definition itself __MUST__ contain information that can be used to locate it directly. We will use that
+ information in the [Component](4.component.md) to refer to the workload schema instead of using the name of the workload definition. 
+
+We recommend that the `name` here follows the format described in [the Group, Version, Kind section of part 2](2.overview_and_terminology.md).
+
+We also recommend that the `name` of the workload (in the metadata section) to include the Group and Kind of the schema.
```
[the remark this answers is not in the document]
resouer: I'd strongly suggest to include samples instead of jumping to some other place. The current narrative is not easy to understand even from my view.

3.workload.md
```diff
@@ -0,0 +1,378 @@
...
+
+## Workload Kinds
```
resouer: Kinds have special meaning in K8s world. Let's name it "Workload Ladder" or sth fancy.

3.workload.md
```diff
@@ -0,0 +1,378 @@
...
+
+There are three kinds of workloads:
+
+__Core workloads__ are the kind of workloads that every OAM runtime MUST implement according to spec definition
+
+__Standard workloads__ are the kind of workloads that a OAM runtime can choose to implement. However, the
+ implementation MUST strictly adhere to the schematics defined in the spec 
+
+__Extended workloads__ OAM platforms are free to define extra workload kinds and
```
resouer: and ... ?

3.workload.md
```diff
@@ -0,0 +1,378 @@
...
+ 
+### Extended Workload Kind
+An extended workload MUST be in the `extended.oam.dev` group.
+
+Extended workload kind are _per runtime_, meaning that each runtime may define its own extended workloads schema beyond this specification. 
+
+## Core Workload
+The following core workload kinds are defined by this specification
+
+### Container Workload
```
resouer: I strongly suggest to move spec of core workloads (the whole section below) to a separate directory (e.g. `spec/core_workloads/container_workload.md` etc), and we will have more core workloads in the future.

My opinion is they are not part of OAM spec, they just allowed to stay in OAM spec repo because they are "core".

2.overview_and_terminology.md
ryanzhang-oss: I don't follow.  The component.md is in this PR.

3.workload.md
ryanzhang-oss: This only happens once.  Also, I thought  that our GKV are "inspired" by the K8s but  not direct dependent on K8s definition.

3.workload.md
ryanzhang-oss: Same as above. Btw, this section is copied from the existing spec

3.workload.md
negz: Strongly agree. :) Though we might consider linking instead to the API conventions document; I believe this is the closest thing that exists to a formal specification of the Kubernetes API.

https://github.com/kubernetes/community/blob/1d6d124/contributors/devel/sig-architecture/api-conventions.md

3.workload.md
negz: > Note that it could be simply an indexing for a schema.

It's not immediately obvious to me what this is intending to communicate. Could we word it more clearly?

3.workload.md
negz: > Must be formatted in a way that can uniquely identify a specific version of the definition

Cross-linking this discussion; https://github.com/crossplane/crossplane/pull/1276#pullrequestreview-362313530. I'm really not sure requiring a specific version makes sense in the Kubernetes world.

3.workload.md
```diff
@@ -0,0 +1,378 @@
...
+DefinitionRef defines where to find the workload schema
+
+| Attribute | Type | Required | Default Value | Description |
+|-----------|------|----------|---------------|-------------|
+| `name` | `string` | Y | | The workload schema's name. Must be formatted in a way that can uniquely identify a specific version of the definition  |
+
+The workload definition itself __MUST__ contain information that can be used to locate it directly. We will use that
+ information in the [Component](4.component.md) to refer to the workload schema instead of using the name of the workload definition. 
+
+We recommend that the `name` here follows the format described in [the Group, Version, Kind section of part 2](2.overview_and_terminology.md).
```
negz: Is this suggesting, for example, `name: local.dev/v7alpha2.Proxy`? I recommend we do not use a field named `name` unless we're actually referring to the `metadata.name` of some other Kubernetes object, and I don't believe this is a valid Kubernetes object name (per https://kubernetes.io/docs/concepts/overview/working-with-objects/names/).

My thinking here was that `name` would refer to a named instance of a `CustomResourceDefinition`. If we instead want to simply refer to a particular kind of resource we should use `apiVersion` and `kind` (and not `name`).

3.workload.md
```diff
@@ -0,0 +1,378 @@
...
+### Standard Workload Kind
+
+A standard workload MUST be in the `standard.oam.dev` group.
+
+A standard workload MUST be implemented according to the definitions in this section if any runtime chooses to implement it.
+ 
+We do not have any standard workload defined in the current spec. We may introduce new workloads or promote extended workloads to standard if there is a consensus in the community.
+ 
+### Extended Workload Kind
+An extended workload MUST be in the `extended.oam.dev` group.
```
negz: Part of our goal with this spec update was to allow arbitrary Kubernetes CRs to act as extended workloads. This isn't possible if all extended workloads MUST be in the `extended.oam.dev` API group. I don't think we should have an opinion about the group name of extended workloads.

3.workload.md
```diff
@@ -0,0 +1,378 @@
...
+A containerized workload is used for long-running workloads. Here is the definition of the containerized workload
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: workloadDefinition
+metadata:
+  name: containerWorkload.core.oam.dev
+spec:
+  definitionRef:
+    name: core.oam.dev/v1alpha2.containerWorkload
```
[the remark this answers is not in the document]
negz: I believe `core.oam.dev/v1alpha2.ContainerWorkload` is not a valid `metadata.name` (due to the `/` and the uppercase characters) and thus should not be the value of a field named `name`. If what we're trying to do is reference a particular GVK (not a named instance of, for example, a CRD) then we should write:

```yaml
definitionRef:
  kind: ContainerisedWorkload
```

2.overview_and_terminology.md
ryanzhang-oss: What does `type` mean here? Can you give a type example?

3.workload.md
ryanzhang-oss: > Put otherwise, I agree that we should explain that `WorkloadDefinition` is intended for use by infrastructure operators, but I do not think we should go so far as to say application operators MUST NOT be allowed to create workload kinds.

Agree,  this is copied from the existing spec (https://github.com/oam-dev/spec/blob/master/7.workload_types.md#7-workload-types) so I didn't change it as this PR is just about the schema change.  I didn't know why we put such a strong statement in the begining. We can add a PR to resolve this.

3.workload.md
ryanzhang-oss: would it make sense in a non-k8s world?

3.workload.md
ryanzhang-oss: >  This isn't possible if all extended workloads MUST be in the `extended.oam.dev` API group.

Since the CRDs' group name does not have to be the same as its parent workloadDefinition's group name, I wonder why it is not possible to refer to any CRDs?

3.workload.md
ryanzhang-oss: I thought about that too, didn't know where to put them.

2.overview_and_terminology.md
resouer: type: networking

type: healthgroup

type: securitygroup

2.overview_and_terminology.md
resouer: Oh I see, I missed that file.

3.workload.md
resouer: I'd suggest we change the narrative in this PR :-)

3.workload.md
negz: Perhaps I'm misunderstanding something here. This reads to me like you're saying any extended workload, i.e. any kind blessed as a workload via a `WorkloadDefinition` must be in the `extended.oam.dev` group. So for example I could not use a Crossplane `RedisCluster` as a workload, because it's in `cache.crossplane.io`, not `extended.oam.dev`:

```yaml
kind: Component
metadata:
  name: example-cache
spec:
  workload:
    # Are you saying this MUST be extended.oam.dev?
    apiVersion: cache.crossplane.io/v1alpha1
    kind: RedisCluster
    spec:
      engineVersion: "3.2"
```

3.workload.md
resouer: Yes, the api-conventions.md doc is the best candidate here. It also reminds me that referring to an epoch design doc of Kube feels much better than linking K8s doc. Good point!

To @ryanzhang-oss , for metadata and resource model, I'd suggest we directly use it instead of "inspired". "Inspire" is not good unless we significantly improved it and decide not contribute back. 

If needed (seems not IMO), we can clearly clarify that we only support a certain subset of the whole k8s resource model convention. But I'd strongly prefer eliminating the feeling of "inspire".

3.workload.md
resouer: This doc is better: https://github.com/kubernetes/community/blob/1d6d124/contributors/devel/sig-architecture/api-conventions.md#metadata

My opinion is we need to have a clearer relationship between overview_terminology.md section and K8s resource model than today. /cc @negz 

We can fix this in following up PR of course.

3.workload.md
resouer: I am in favor of Ryan's proposal:

```yaml
name: containerisedworkload.core.oam.dev
version: v1beta1
```
The version field should be optional of course (only needed when multiple version CRDs are installed which is a rare case).

4.component.md
```diff
@@ -74,415 +47,21 @@ The Parameters section defines all of the configurable parameters for this compo
...
-
-Our first component is a frontend administrative interface. End users (bot
-admins) authenticate using basic authentication. The interface permits end users
-to configure what kind of content the bot will tweet and the intervals at which
-it will do so.
-
-Because this component is stateless, horizontally scalable, and exposes an endpoint, it is best described
-as a `Server`. The username, password, and address of the backend
-component are configurable.
+This section illustrates use of the component schematic defined above to describe a component.
```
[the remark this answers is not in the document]
ryanzhang-oss: the "component  schematic" means the schema of a component.

4.component.md
```diff
@@ -1,54 +1,27 @@
...
-- `5M` is 5 megabytes
-- `7G` is 7 gigabytes
-- `100T` is 100 terabytes
-- `9999P` is 9999 petabytes
-
-If a `B` is appended after the unit letter, it MUST be ignored. Thus, `5M` and `5MB` are treated as identical. Case is unimportant. `15k` and `15K` MUST be treated as the same value.
-
 ## Representation
 
 All component schematics are canonically represented as JSON and can be validated using the provided JSON Schema. Because it is more amenable for human consumption, YAML is also supported, and most of the examples are given in YAML. For the sake of validation, YAML may be converted to an equivalent JSON representation before validating it against the JSON Schema.
```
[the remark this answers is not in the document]
ryanzhang-oss: A component can't be validated by the workload definition. The workload property part of the component schematic can be validated against the corresponding workloadDefinition schema. Right?

3.workload.md
resouer: I think it's fine to have `name` (and it's name of CRD) in xxxDefinition, but in other obj, I agree with your point.

3.workload.md
resouer: @negz  I think Ryan means the WorkloadDefinition under `extended.oam.dev` group instead of `core.oam.dev`? It should be mentioned in xxxDefinition section.

3.workload.md
resouer: Let's do `spec/core_workloads/container_workload.md`. /cc @negz

3.workload.md
ryanzhang-oss: > Also I'm not sure that "indexing" works for non-CRD schema. I'm thinking maybe:

it seems that it should work as long as the "name" is unique? Is it hard to do? With the schematic and name together, what does the "name" mean here?

4.component.md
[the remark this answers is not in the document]
wonderflow: Yeah, you are right

3.workload.md
wonderflow: Yeah, the description here has some mislead.  It is workloadDefinition MUST be in the `extended.oam.dev` group not workload itself.

negz: Does this mean we'll have multiple `kind: WorkloadDefinition`, each in a different API group (`extended.oam.dev`, `core.oam.dev`, etc?). I imagine core workloads will still require a `WorkloadDefinition`.

3.workload.md
negz: Happy for it to move out of this doc. I suggest calling it a "_containerised_ workload" rather than a "container workload" though.

3.workload.md
wonderflow: Sorry @negz ,  I'm wrong.  I think we didn't need `extended.oam.dev` group at all.  I think the group `core.oam.dev` means they are absolutely defined in OAM Spec and every OAM runtime MUST follow it.  The group `standard.oam.dev` means they are defined in OAM Spec and can be optionally implemented by runtime.

3.workload.md
vturecek: I'm hesitant about linking directly to the Kubernetes docs for formal definitions, including structure or field names. If Kubernetes decides to suddenly change then name or usage of a field, we would need to first assess that change and decide if it should be applied here too. Conceptual docs are probably fine since a change to the wording of a concept in the Kubernetes doc won't necessarily break an OAM runtime.

3.workload.md
```diff
@@ -0,0 +1,99 @@
...
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: WorkloadDefinition
+metadata:
+  name: schema.example.oam.dev
+spec:
+  definitionRef:
+    schemaName: schema.example.oam.dev
+```
```
vturecek: the example here doesn't follow the recommendation we just gave above to use GVK. Is it enough to simply recommend GVK, or should it be required?

3.workload.md
vturecek: This also contradicts the recommendation above, which says that we recommend the workload definition name match the schema name. For extended workload types, that will be impossible if someone is adding existing workloads, like the RedisCluster in @negz 's example. 

I don't think we need the `extended.oam.dev` group at all here.

core/workloads/containerized_workload/containerized_workload.md
```diff
@@ -0,0 +1,288 @@
+# Containerized Workload
+A containerized workload is used for long-running workloads in containers. Here is the definition of the containerized workload.
```
vturecek: This is where execution policies are needed. A containerized workload need not strictly be a long-running workload.

3.workload.md
ryanzhang-oss: `containerized` or `containerised`?

3.workload.md
ryanzhang-oss: We didn't recommend GVK as the result of several discussions says no "V" needed. I will modify the wording a bit.

core/workloads/containerized_workload/containerized_workload.md
ryanzhang-oss: IMO, there are two ways to solve this.

- One is to rename this to be `containerized server` and add another workload called `containerized task` which has a different schema (task specific) from server. 

- The other is to keep this as the only core containerized workload and add policies for components.

These two ways can actually co-exist too.

3.workload.md
ryanzhang-oss: > ```yaml
>   workload:
>     # Are you saying this MUST be extended.oam.dev?
>     apiVersion: cache.crossplane.io/v1alpha1
>     kind: RedisCluster
>     spec:
>       engineVersion: "3.2"
> ```
No, this is the apiVersion of the CRD. The extended workloadDefinition itself needs to be in the "extended.oam.dev". The CRD it refers to can be anything.

3.workload.md
resouer: @vturecek Actually, `.../generated/kubernetes-api/{version}` is the permanent link of versioned K8s api spec. It's much safer than what OAM spec currently linked in the doc, e.g.: https://github.com/oam-dev/spec/blob/master/2.overview_and_terminology.md#label-format which only reflects the latest version of K8s api.

Also, I propose we directly refer K8s metadata instead of creating a "simplified" version, because in practice, the OAM K8s runtime still need to use full version of [k8s metadata](https://github.com/oam-dev/rudr/blob/319448fb4cfd970152ecb24d3722536d66ea6bd7/examples/components.yaml#L3). The OAM spec version of "name + labels + annotations" actually is not enough to work at all. In that sense, current rudr doesn't fully follow OAM spec.

I'm totally OK we fix it in following PR but we need to fix it ASAP indeed.

resouer: What I proposed above also applies to GVK section and many places in "overview and terminology" doc . Non-k8s impl will also be benefit a lot since it could import K8s types directly in many languages. It doesn't make sense to create sth "similar" in OAM spec.

3.workload.md
zhxu2: > workload:
>     # Are you saying this MUST be extended.oam.dev?
>     apiVersion: cache.crossplane.io/v1alpha1
>     kind: RedisCluster
>     spec:
>       engineVersion: "3.2"
> 
> No, this is the apiVersion of the CRD. The extended workloadDefinition itself needs to be in the "extended.oam.dev". The CRD it refers to can be anything.

meaning
```
apiVersion: **extended**.oam.dev/v1alpha1
kind: WorkloadDefinition
metadata:
  name: rediscluster.cache.crossplane.io
spec:
  definitionRef:
    name: rediscluster.cache.crossplane.io
```
?

2.overview_and_terminology.md
resouer: After think it a while I feel in most cases Kind is enough, let's keep it as is.

3.workload.md
ryanzhang-oss: > apiVersion: **extended**.oam.dev/v1alpha1
> kind: WorkloadDefinition
> metadata:
>   name: rediscluster.cache.crossplane.io
> spec:
>   definitionRef:
>     name: rediscluster.cache.crossplane.io

yeah,  is this ok?

hongchaodeng: > allow arbitrary Kubernetes CRs to act as extended workloads.

I totally agree on this. OAM is supposed to manage, not replace CRDs. The "extended" needs some context that needs some explanation:

When we started the project, our vision is to standardize application management. There are levels of standardization in our roadmap: _core_ for minimum, _standard_ for cloud ubiquitous, and _extended_ for things that are popular but not ubiquitous. We host and govern those features/CRDs/Operators in OAM to build a standardized application delivery experience.

This is not a technical issue, but more a community building effort.

3.workload.md
zhxu2: > I think it's fine to have name (and it's name of CRD) in xxxDefinition, but in other obj, I agree with your point.

it seems DefinitionRef Name here is a GVK string. Yes we can find the referred CR instance by parsing it but the combination of name: GVK as Nic suggested, is not very intuitive

3.workload.md
resouer: Allow me summarize a bit, so Ryan proposed is:

For core workloads, it looks like:
```yaml
kind: WorkloadDefinition
metadata:
  name: x
spec:
  definitionRef:
    name: x.core.oam.dev
```
For standard workload, it will be:

```yaml
apiVersion: standard.oam.dev/v1alpha1
kind: WorkloadDefinition
metadata:
  name: y.database.alibaba.com
spec:
  definitionRef:
    name: y.database.alibaba.com
```

For extended workload, it will be:

```yaml
apiVersion: extended.oam.dev/v1alpha1
kind: WorkloadDefinition
metadata:
  name: z.cache.crossplane.io
spec:
  definitionRef:
    name: z.cache.crossplane.io
```
Am I right?

ryanzhang-oss: Here is what I think makes sense.  WorkloadDefinition always belongs to core.oam.dev as it's OAM's core concept. The apiground in the schema for core workload is core.oam.dev.  The apiground in the schema for standard workload is standard.oam.dev. There is no restriction on the api-group for extended.

For core workloads, it looks like:
```yaml
kind: WorkloadDefinition
metadata:
  name: x.core.oam.dev
spec:
  definitionRef:
    name: x.core.oam.dev
```
For standard workload, it will be:
```yaml
kind: WorkloadDefinition
metadata:
  name: y.standard.oam.dev
spec:
  definitionRef:
    name: y.standard.oam.dev
```
For extended workload, it will be:
```yaml
kind: WorkloadDefinition
metadata:
  name: z.cache.crossplane.io
spec:
  definitionRef:
    name: z.cache.crossplane.io
```

3.workload.md
hongchaodeng: Hi @vturecek 

Please see referenced discussions 
- https://github.com/oam-dev/spec/pull/316#discussion_r383492681
- https://github.com/crossplane/crossplane/pull/1276#discussion_r382321173 (@negz 's PR in crossplane)

3.workload.md
```diff
@@ -0,0 +1,99 @@
...
+ 
+### Extended Workload
+An extended workload MUST be in the `extended.oam.dev` group.
+
+Extended workload kind are _per runtime_, meaning that each runtime may define its own extended workloads schema beyond this specification. 
+
+## Core Workload definitions
+The following core workloads are defined in our current OAM specification.
+
+### Containerize Workload
```
negz:
```suggestion
### Containerized Workload
```

core/workloads/containerized_workload/containerized_workload.md
negz: > which has a different schema (task specific) from server

How do we expect task specific schema would differ? Under the contemporary spec the schema of task and server (and worker) are identical.

core/workloads/containerized_workload/containerized_workload.md
```diff
@@ -0,0 +1,288 @@
...
+apiVersion: core.oam.dev/v1alpha2
+kind: WorkloadDefinition
+metadata:
+  name: containerizedworkload.core.oam.dev
+spec:
+  definitionRef:
+    schemaName: containerizedworkload.core.oam.dev
+```
+
+Here is the schematics of the containerized workload.
```
negz:
```suggestion
Here is the schematic of a containerized workload. 
```

core/workloads/containerized_workload/containerized_workload_schema.json
```diff
@@ -0,0 +1,361 @@
+{
```
negz: Cross-linking https://github.com/oam-dev/spec/pull/319/files#r384207097. How would we feel about standardising on YAML OpenAPI v3 schemas instead of JSON schemas?

3.workload.md
negz: This is very hard to say an an Australian, but since software tends to use US english it should probably be containeri**z**ed. 😢

3.workload.md
negz: I think I had a stale view of this thread when I wrote https://github.com/oam-dev/spec/pull/316#discussion_r384223831. It sounds like I agree with @ryanzhang-oss's comment at https://github.com/oam-dev/spec/pull/316#discussion_r384219536.

2.overview_and_terminology.md
```diff
@@ -89,9 +89,8 @@ An application configuration has the following parts:
 
 ## Representing OAM Objects as Schematics
 
-The OAM specification represents OAM objects (workloads, components, trait definitions, application configurations
-, etc
-.) as _schematics_. A schematic is a structured document that provides a declaration of an object or an object's desired state. Throughout this specification, schematics are represented in [YAML](https://yaml.org/). However, nothing in this specification forecloses the possibility of representing schematics as JSON documents or other similarly structured textual or binary representations.
+The OAM specification represents OAM objects (workloads, components, trait definitions, application configurations, etc.) as _schematics_.
+A schematic is a structured document that provides a declaration of an object or an object's desired state. Throughout this specification, schematics are represented in [YAML](https://yaml.org/). However, nothing in this specification forecloses the possibility of representing schematics as JSON documents or other similarly structured textual or binary representations.
```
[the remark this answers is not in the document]
ryanzhang-oss: It seems that the googling worked.

3.workload.md
```diff
@@ -0,0 +1,99 @@
+# 3. Workload
+
+This section defines the workload definition schematics.
+
+Usually, workload definitions are provided by the OAM runtime (just as traits are) so that application developers may inspect the OAM runtime and learn what workloads are available for use.  
+
+## Top-Level Attributes
+
+The following attributes are common across all schemata defined in this document. They follow the [Kubernetes specification](https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/#required-fields).
```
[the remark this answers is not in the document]
ryanzhang-oss: I think MSFT/Alibaba/Upbound should find time this week to discuss this. I think we should address this in a follow-up PR since this is only getting into alpha2 branch. This PR is more for the technical part.

3.workload.md
```diff
@@ -0,0 +1,99 @@
...
+#### DefinitionRef
+
+DefinitionRef defines an index to find the workload schema.
+
+| Attribute | Type | Required | Default Value | Description |
+|-----------|------|----------|---------------|-------------|
+| `schemaName` | `string` | Y | | The workload schema's name. |
+
+The workload schema definition itself __MUST__ contain information that can be used to uniquely identify it. OAM runtime will use that information in a [Component](4.component.md) to refer to the workload schema instead of using the name of the workload definition.
+ We recommend to use the format [Group/Version/Kind](2.overview_and_terminology.md#Group) described in section 2 as the unique identification in the schema.
```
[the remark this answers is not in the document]
ryanzhang-oss: > ```yaml
> spec:
>   definitionRef:
>     name: core.oam.dev/v1alpha1.ContainerizedWorkload
> ```
> 
> Is this recommending that runtimes use the above `group/version.Kind` syntax to uniquely identify a schema? I think this would be surprising and misleading to folks familiar with Kubernetes, and make implementing the spec a little more difficult for runtimes that are built using Kubernetes.
> 
> Quoting API conventions:
> 
> > The name of a field referring to another resource of kind Foo by name should be called fooName. The name of a field referring to another resource of kind Foo by ObjectReference (or subset thereof) should be called fooRef.
> 
> > Object references should either be called fooName if referring to an object of kind Foo by just the name (within the current namespace, if a namespaced resource), or should be called fooRef, and should contain a subset of the fields of the ObjectReference type.
> 
> i.e. `spec.fooRef.name` should refer to the `metadata.name` of a `kind: Foo`. In the case of `definitionRef` we omit the kind and leave it up to the runtime (perhaps it's a `kind: CustomResourceDefinition`, perhaps it's a `kind: SweetXMLSchema`) but for Kubernetes runtimes I think this should be the _name_ of the definition, not a GVK. The `group/version.Kind` syntax is not a valid Kubernetes name (due to the `/` and uppercase kind), and CRDs specifically [MUST](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition) be named plural.group, e.g. `containerizedworkloads.oam.core.dev`.
> 
> So TL;DR I expect all runtimes that are implemented atop the Kubernetes API server would need to write:
> 
> ```yaml
> spec:
>   definitionRef:
>     name: containerisedworkloads.oam.core.dev
> ```
No, that link actually points to exactly what you describe here

3.workload.md
```diff
@@ -0,0 +1,99 @@
...
+A standard workload MUST be in the `standard.oam.dev` group.
+
+A standard workload MUST be implemented according to the definitions in this section if any runtime chooses to implement it.
+ 
+We do not have any standard workload defined in the current spec. We may introduce new workloads or promote extended workloads to standard if there is a consensus in the community.
+ 
+### Extended Workload
+An extended workload MUST be in the `extended.oam.dev` group.
+
+Extended workload kind are _per runtime_, meaning that each runtime may define its own extended workloads schema beyond this specification.
```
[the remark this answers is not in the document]
ryanzhang-oss: Agree,  I took a shoot at re-wording it,  not my forte...

3.workload.md
vturecek: So are we saying that OAM-compliant runtimes must implement the full Kubernetes metadata object model, as specified in the Kubernetes API conventions doc? That seems extreme to me. Namespaces would be required by every runtime even though that has never been a concept we've had in OAM and to me seems outside of the scope of an application modeling definition.

As I stated earlier, I'm fine linking off to _concepts_ where we feel they are identical. If the content is versioned, all the better. 

I liked the suggestion earlier that we clearly label parts of the spec with the role they are most relevant to (app developer, app operator, infra operator). But I'm not sure what it means to have a separate part of the spec for platform builders, or what we gain from that. If they are split then I'm assuming there are object models we define for platform builders that end users don't see, in which case I'm not sure why we even care about those.

core/workloads/containerized_workload/containerized_workload.md
vturecek: I don't think they would differ at all. That's the problem we solved with workloadType in OAM alpha1 and have now re-introduced here. Rather than define multiple workload definitions with identical schemas, I think we need a way to classify execution behavior or something like that, so that we avoid duplicate schema definitions and have a way to express whether or not a trait is applicable to a given component.

core/workloads/containerized_workload/containerized_workload.md
```diff
@@ -0,0 +1,288 @@
...
+A containerized workload is used for long-running workloads in containers. Here is the definition of the containerized workload.
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: WorkloadDefinition
+metadata:
+  name: containerizedworkload.core.oam.dev
+spec:
+  definitionRef:
+    schemaName: containerizedworkload.core.oam.dev
```
negz:
```suggestion
    name: containerizedworkloads.core.oam.dev
```

core/workloads/containerized_workload/containerized_workload.md
```diff
@@ -0,0 +1,288 @@
+# Containerized Workload
+A containerized workload is used for long-running workloads in containers. Here is the definition of the containerized workload.
+
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: WorkloadDefinition
+metadata:
+  name: containerizedworkload.core.oam.dev
```
negz:
```suggestion
  name: containerizedworkloads.core.oam.dev
```

core/workloads/containerized_workload/containerized_workload.md
```diff
@@ -0,0 +1,288 @@
...
+Here is the schematics of the containerized workload. 
+
+### Top-Level Attributes of a containerized workload 
+
+These attributes provide top-level information about the containerized workload.
+
+| Attribute | Type | Required | Default Value | Description |
+|-----------|------|----------|---------------|-------------|
+| `apiVersion` | `string` | Y ||  `core.oam.dev/v1alpha2` |
+| `kind` | `string` | Y ||  `containerWorkload` |
```
negz:
```suggestion
| `kind` | `string` | Y ||  `ContainerizedWorkload` |
```

core/workloads/containerized_workload/containerized_workload.md
negz: @vturecek To be honest it seems pretty reasonable to me for workloads that are conceptually different things to encode that difference in their GVK, even if they have identical schema. Previously we effectively had a `ComponentSchematic` type with a subtype encoded in `spec.workloadType`. Now we have a `Component` type with a subtype encoded in `spec.workload.kind`.

I could see a case for moving the encoding of whether or not something was a singleton into the schema as a field under `spec` and simply having three kinds with identical schemas; probably `ContainerizedServer`, `ContainerizedWorker`, and `ContainerizedTask` similar to what @ryanzhang-oss suggested.

core/workloads/containerized_workload/containerized_workload.md
hasheddan:
```suggestion
    name: containerizedworkload.core.oam.dev
```
per: https://github.com/crossplane/crossplane/pull/1276#discussion_r383647996

hasheddan: See this was already pointed out while I was reviewing, resolving in favor of https://github.com/oam-dev/spec/pull/316#discussion_r384720593

3.workload.md
resouer: @vturecek I also agree directly referring Kubernetes design doc is a good start to make the spec doc simpler.

While after digesting your comment above, I feel that the OAM spec doc cares very much about end users. But honestly, my personal observation is the current spec is barely friendly to them. I'd guess it's also the main reason that OAM spec was commented as "complex":  it reads like a doc for someone to implement an OAM platform rather than material to grape OAM concepts immediately at first glance, or more possibly, those two are inextricably mixed together in the spec.

In that sense, it's exactly the same mistake that K8s Deployment is making and we claim to solve in OAM :-)

Another concern is, the current spec is not implementers friendly either. For example, from reading the spec, I have feeling that we are suggesting implementers to code OAM metadata types from scratch. I really doubt there's any implementer will do that. /cc @negz 

So I don't mind if it's named "splitting", but I'd highlight the idea like: OAM spec **should also consider separate concerns** in its content, specifically:

1. straightforward docs for end users to get the core ideas of OAM in 10 min instead of reading a 10 pages specification. IMO, detailed fields of OAM metadata is not part of this. 
2. detailed and accurate docs for implementers to implement OAM on whatever runtime, and what's more important, with minimal effort. For example, again the `metadata` section, the message will be much clearer like "OAM metadata is a subset of K8s metadata (link) and here're the fields OAM cares".

I don't see any reason that we can't achieve this or have to mix them in one spec doc. Though I'd suggest we move this discussion out of this PR, it need to be fixed by following ups (and would be better to involve a tech writer) /cc @ryanzhang-oss

4.component.md
```diff
@@ -74,573 +48,72 @@ The Parameters section defines all of the configurable parameters for this compo
...
+    kind: ContainerizedWorkload
+    metadata:
+      name: sample-workload
+    spec:
+      osType: linux
+      containers:
+      - name: my-cool-workload
+        image:
+          name: example/very-cool-workload:0.1.2
+          digest: sha256:verytrustworthyhash
```
hasheddan:
```suggestion
        image: example/very-cool-workload:0.1.2@sha256:verytrustworthyhash
```
Kubernetes uses [docker syntax](https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier#pull-an-image-by-digest-immutable-identifier) for pulling images by digest.

core/workloads/containerized_workload/containerized_workload.md
ryanzhang-oss: why using purls? is it K8s convention? We want to keep the two "names" the same.  Does this mean that the CRD name needs purls too?

core/workloads/containerized_workload/containerized_workload_schema.json
ryanzhang-oss: Let's discuss this and the k8s link problem in another PR.

4.component.md
```diff
@@ -74,573 +48,70 @@ The Parameters section defines all of the configurable parameters for this compo
...
+            path: /healthz
+        readinessProbe:
+          httpGet:
+            port: 8080
+            path: /healthz
+  parameters: 
+  - name: imageName
+    required: false
+    fieldPaths: 
+    - "spec.containers[0].image.name"
```
hasheddan:
```suggestion
    - "spec.containers[0].image"
```

core/workloads/containerized_workload/containerized_workload.md
negz: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition

> name must match the spec fields below, and be in the form: <plural>.<group>

I agree that we want to keep the two names the same, and we're subject to a Kubernetes constraint here. Kubernetes requires that the CRD defining `kind: ContainerizedWorkload` in `group: core.oam.dev` be named `containerizedworkloads.core.oam.dev` (plural). So if we want to refer to a CRD by name (and thus make getting the CRD way easier for our controllers) we must use the plural name.

core/workloads/containerized_workload/containerized_workload.md
ryanzhang-oss: > simply having three kinds with identical schemas; probably `ContainerizedServer`, `ContainerizedWorker`, and `ContainerizedTask` similar to what @ryanzhang-oss suggested.

Since those are all core workloads, we need to spell out their schematic in the spec. It seems we either 
- have 3 identical schemas in the spec 
- explain that we use the schema's name to differentiate the runtime characteristics of a workload from another workload with an identical schema.

I think the second approach is pretty much the same as in spec alphav1 as @negz says. I am just not sure how can we evolve it to the policy approach which I think we all agree is the right direction.

Let's find a time to discuss this in a meeting.

3.workload.md
```diff
@@ -0,0 +1,128 @@
...
+metadata:
+  name: x.standard.oam.dev
+spec:
+  definitionRef:
+    name: x.standard.oam.dev
+```
+
+A standard workload MUST be implemented according to the definitions in the spec if any runtime chooses to implement it.
+ 
+We do not have any standard workload defined in the current spec. We may introduce new workloads or promote extended workloads to standard if there is a consensus in the community.
```
resouer: Can we make a clear claim that standard workloads will be promoted from extended workloads? I'm hoping users can get the point of "promotion ladder" in this section.

6.traits.md
```diff
@@ -1,4 +1,5 @@
-# 5. Traits
+# 6. Traits
+**Target User: Application Operator**
```
resouer: I'd prefer: **Target Audience: ...**, i.e. the reader of the doc.

3.workload.md
ryanzhang-oss: I am not sure if the standard workloads have to be promoted from an extended one. It can be a new one that is similar to a bunch of extended workloads. The core one has to be promoted from standard as we won't change its definition.

5.application_scopes.md
```diff
@@ -1,4 +1,5 @@
-# 4. Application Scopes
+# 5. Application Scopes
+**Target User: Application Operator**
```
hongchaodeng: Aren't both Scope and Trait defined by Infra Ops? They define the spec, and app ops use it.

3.workload.md
resouer: Emm, can we highlight the idea of promotion ladder somewhere? It seems implicit for now.

5.application_scopes.md
resouer: It's another reason I begin to prefer "Target Audiences", both Infra Ops and App Ops are audiences for this doc in my mind. In the future we could cut a clearer boundary.

4.component.md
```diff
@@ -74,573 +48,70 @@ The Parameters section defines all of the configurable parameters for this compo
...
-The format of the path is specific to the operating system of the consuming component, though implementations SHOULD provide support for UNIX-like path representations.
-
-| Attribute | Type | Required | Default Value | Description |
-|-----------|------|----------|---------------|-------------|
-| `name` | `string` | Y |  | Specifies the name used to reference the path. |
-| `mountPath` | `string` | Y |  | Specifies the actual mount path in the filesystem. |
-| `accessMode` | `string` | N | `RW` | Specifies the access mode. Allowed values are `RW` (read/write) and `RO` (read-only). |
-| `sharingPolicy` | `string` | N | `Exclusive` | The sharing policy for the mount, indicating if it is expected to be shared or not. Allowed values are `Exclusive` and `Shared`. |
-| `disk` | [`Disk`](#disk) | N |  | Specifies the attributes of the underneath disk resources required by the volume. |
+| `FieldPaths` | `[]string` | Y | | JSON field paths. |
```
wonderflow:
```suggestion
| `fieldPaths` | `[]string` | Y | | JSON field paths. |
```

4.component.md
```diff
@@ -74,573 +48,70 @@ The Parameters section defines all of the configurable parameters for this compo
...
-
-### TCPSocket
-
-| Attribute | Type | Required | Default Value | Description |
-|-----------|------|----------|---------------|-------------|
-| `port` | `integer` | Y || The TCP socket within the container that should be probed to assess container health. |
-
-Port must be an integer value greater than `0`.
+Parameter `FieldPaths` specifies an array of fields within this Component's workload that will be overwritten by the
+ value of this parameter. `FieldPaths` are specified as JSON field paths without a leading dot, for example `spec.containers[0].image`.
```
wonderflow:
```suggestion
Parameter `fieldPaths` specifies an array of fields within this Component's workload that will be overwritten by the
 value of this parameter. `fieldPaths` are specified as JSON field paths without a leading dot, for example `spec.containers[0].image`.
```

core/workloads/containerized_workload/containerized_workload_schema.json
wonderflow: I agree we should use YAML OpenAPI v3 schemas.

More than that, this schema is out of date, as it follows the old version of OAM spec.

4.component.md
```diff
@@ -1,54 +1,28 @@
...
-If a `B` is appended after the unit letter, it MUST be ignored. Thus, `5M` and `5MB` are treated as identical. Case is unimportant. `15k` and `15K` MUST be treated as the same value.
-
 ## Representation
 
 All component schematics are canonically represented as JSON and can be validated using the provided JSON Schema. Because it is more amenable for human consumption, YAML is also supported, and most of the examples are given in YAML. For the sake of validation, YAML may be converted to an equivalent JSON representation before validating it against the JSON Schema.
 
 ### Top-Level Attributes
 
-The following attributes are common across all schemata defined in this document. This structure is designed to allow object type to be determined by scanning two fields (`apiVersion` and `kind`), and provide access to common fields in the `metadata` section. The `spec` section is where type-specific information is located, and in this section the `spec` section defines a `Component`.
+These attributes provide top-level information about the component definition.
```
wonderflow: I'm not quite understand why this attributes is about **component definition** not **component**

wonderflow:
LGTM！Great work！

hongchaodeng:
Since @negz also approves this, let's merge this and move forward. We can fix the minor issues later.

PR #319, opened 2020-02-24
[1.0.0-alpha2] application scopes and traits

wonderflow:
Revamped application scopes and traits part mainly according to #304


5.application_scopes.md
```diff
@@ -65,28 +65,22 @@ The spec defines the constituent parts of a scope.
...
+
+A scope definitionRef __MUST__ must be unique by `name` and `version`. The `schematic` field is optional as the definitionRef could be simply an indexing for a schema. In this case a runtime itself __SHOULD__ store definitions and a definition __MUST__ contain a schematic which can validate scope spec.
 
-The `name` field allows Unicode letters, numbers and `_` and `-`. The `description` field allows those characters plus whitespace and punctuation characters.
 
 ## Deployment
-Application scope instances are defined and deployed in an application configuration. See [Application Configuration](7.application_configuration.md) for more information on deploying scopes.
+Application scope instances are defined and deployed by their standalone way.
 
 ## Core scope type definitions
```
ryanzhang-oss: Can we also move all the definitions of core scopes to a separate place?

6.traits.md
```diff
@@ -53,7 +53,7 @@ Implementations of an OAM runtime SHOULD make all supported traits discoverable
```
ryanzhang-oss: We no longer have workload `type` anymore.  I think we need to modify the spec temporarily until we introduce `type` back.

6.traits.md
```diff
@@ -62,26 +62,29 @@ The top-level attributes of a trait define its metadata, version, kind, and spec
...
+| `name` | `string` | Y | | The name of this trait definition. |
+| `version` | `string` | Y | | The version of this trait definition. |
+| `schematic` | `string` | N | | The schematic of this trait properties. The value is a [JSON schema](https://json-schema.org/). |
 
-The value of the properties field in a trait definition is a [JSON schema](https://json-schema.org/) expressed as json string. When a trait is applied to a component instance, the properties are validated against the schema defined in the trait's `properties` attribute.
+A trait definitionRef __MUST__ must be unique by `name` and `version`. The `schematic` field is optional as the definitionRef could be simply an indexing for a schema. In this case a runtime itself __SHOULD__ store definitions and a definition __MUST__ contain a schematic which can validate trait properties in application configurations.
 
-For example, if a trait defines a schema for properties that requires an array of integers with at least one member, the properties object is expected to be an array of integers with at least one member. During validation of the properties object, the trait's property schema will be applied.
 
 ## Core traits
```
ryanzhang-oss: similarly, can we move the core traits definition out to a different place?

5.application_scopes.md
```diff
@@ -142,48 +165,60 @@ The health scope is defined by the following spec:
...
-      type: double
-      required: false
-    - name: required-healthy-components
-      description: Comma-separated list of names of the components required to be healthy for the scope to be health.
-      type: []string
-      required: false
+  definitionRef:
+    name: healthscope.core.oam.dev
+    version: v1alpha1
+    schematic: |
```
hongchaodeng: According to https://github.com/oam-dev/spec/pull/304/, the scope definitionRef will point to a CRD name. We don't need to embed json schema here, do we?

6.traits.md
```diff
@@ -150,10 +149,9 @@ spec:
   variables:
   components:
     - componentName: frontend
-      instanceName: web-front-end
-      parameterValues:
       traits:
-        - name: manual-scaler
+        - name: manualscalertrait.core.oam.dev
```
hongchaodeng: Probably change this to:
```yaml
trait:
  kind: ManualScaler
  ...
```

5.application_scopes.md
```diff
@@ -104,31 +98,60 @@ The network scope is defined by the following spec:
...
-        If IPs are taken up, operators need to add another subnet into this network.
-      type: string
-      required: Y
-    - name: internet-gateway-type
-      description: The type of the gateway, options are 'public', 'nat'. Empty string means no gateway.
-      type: string
-      required: N
+  allowComponentOverlap: true
+  definitionRef:
+     name: networkscope.core.oam.dev
```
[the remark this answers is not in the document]
wonderflow: yeah， it's a ref same with trait defitionRef

5.application_scopes.md
```diff
@@ -65,28 +65,22 @@ The spec defines the constituent parts of a scope.
...
 | Attribute | Type | Required | Default Value | Description |
 |-----------|------|----------|---------------|-------------|
-| `name` | `string` | Y | | The parameter's name. Must be unique per scope. |
-| `description` | `string` | N | | A description of the parameter. |
-| `type` | `string` | Y | | The parameter's type (One of `string`, `boolean`, `number`). |
-| `required` | `boolean` | N |`false` | Whether a value _must_ be provided for the parameter. |
-| `default` | type indicated by `type` field | N | | The parameter's default value. |
+| `name` | `string` | Y | | The name of this scope definition. |
+| `version` | `string` | Y | | The version of this scope definition. |
+| `schematic` | `string` | N | | The schematic of this scope properties. The value is a [JSON schema](https://json-schema.org/). |
```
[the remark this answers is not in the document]
wonderflow: We just use one, I'll change it

5.application_scopes.md
wonderflow: removed

scopes/core.md
```diff
@@ -0,0 +1,176 @@
...
+metadata:
+  name: networkscope.core.oam.dev
+spec:
+  allowComponentOverlap: true
+  definitionRef:
+     name: networkscope.core.oam.dev
+```
+
+OAM runtime should store the schema with a version, the schema SHOULD be able to validate. For example using JSON Schema like this:
+
```
vturecek: Is this just an example, or is it meant to be the actual definition of a core network scope?

resouer:
Let's hold this PR and focus on #316 first, I believe this one need significant rework after that.

scopes/extended.md
```diff
@@ -0,0 +1,94 @@
+# Extended application scope type definitions
```
vturecek: I think we should remove these entirely. They really only serve as examples and its caused quite a bit of confusion in the past.

traits/core.md
```diff
@@ -0,0 +1,87 @@
...
+  name: manualscalertrait.core.oam.dev
+spec:
+  appliesToWorkloads:
+    - core.oam.dev/v1alpha2.containerWorkload
+  definitionRef:
+    schemaName: manualscalertrait.core.oam.dev
+```
+
+OAM runtime should store the schema with a version, the schema SHOULD be able to validate. For example using JSON Schema like this:
+
```
vturecek: As a core trait, the schema should be part of the specification. Are we saying that it doesn't have to be defined using JSON Schema?

scopes/core.md
```diff
@@ -0,0 +1,176 @@
...
+        }
+      }
+    }
+```
+
+### Instance Example
+
+For example, a network scope instance can be like below, asuming the version is `v1alpha1`:
+
+```yaml
```
ryanzhang-oss: What does this instance mean? We only embedded it in the applicationConfig now, right?

6.traits.md
ryanzhang-oss: The good thing is that we have only 1 workload now so it still works :) 
I think if we can quickly converge on the idea that traits can be applied to a certain "policy" then we can at least put that idea in the spec before the Kubecon demo.

ryanzhang-oss: > Yeah there is a glaring hole here right now.

The hole was there (albeit arguable smaller) already as we didn't define any workload "type" outside of core workloads. One can't find any traits that apply to an extended workload.

6.traits.md
```diff
@@ -53,7 +53,7 @@ Implementations of an OAM runtime SHOULD make all supported traits discoverable
 
 This section is normative because traits are an inspectable (and possibly shareable) part of the system. All traits MUST be representable in the following format.
 
-Traits are defined with schematics like components. Unlike component schematics, however, each trait definition defines its own unique [properties](#properties) object schema to define the configuration options for the trait.
+Traits are defined with schematics like components. Same with workload definition, each trait definition has a definitionRef which can index the object schema to define the configuration options for the trait.
```
ryanzhang-oss: Aren't  traits defined like workload/scope?

5.application_scopes.md
```diff
@@ -65,185 +65,18 @@ The spec defines the constituent parts of a scope.
 
 | Attribute | Type | Required | Default Value | Description |
 |-----------|------|----------|---------------|-------------|
-| `type` | `string` | Y | | Determines type of the application scope. The core types are defined [above](#core-application-scope-types) |
 | `allowComponentOverlap` | `bool` | Y | | Determines whether a component is allowed to be in multiple instances of this scope type simultaneously. When false, the runtime implementation MUST produce an error and stop deployment if an attempt is made to place a component into more than one instance of this scope type simultaneously. |
-| `parameters` | [`[]Parameter`](#parameter) | N | | The scope's configuration options. |
+| `definitionRef` | [`DefinitionRef`](#definitionRef) | Y | | Index of the scope definition. |
 
-The `type` field defines the type name for this type. Examples may be found in the [Core Application Scope Types](#core-application-scope-types) above. Extended scope types MUST NOT be in the `core.oam.dev` namespace.  However, they must follow the [group, version, kind notation](2.overview_and_terminology.md). For example: `scopes.example.com/v1.ExtendedNetworkScope`
+### DefinitionRef
```
ryanzhang-oss: You can probably omit this and put `[DefinitionRef](3.workload.md#DefinitionRef)` at line 69

5.application_scopes.md
```diff
@@ -65,185 +65,18 @@ The spec defines the constituent parts of a scope.
...
-|-----------|------|----------|---------------|-------------|
-| `name` | `string` | Y | | The parameter's name. Must be unique per scope. |
-| `description` | `string` | N | | A description of the parameter. |
-| `type` | `string` | Y | | The parameter's type (One of `string`, `boolean`, `number`). |
-| `required` | `boolean` | N |`false` | Whether a value _must_ be provided for the parameter. |
-| `default` | type indicated by `type` field | N | | The parameter's default value. |
+## Deployment
 
-The `name` field allows Unicode letters, numbers and `_` and `-`. The `description` field allows those characters plus whitespace and punctuation characters.
+Application scope instances are defined and deployed by their standalone way. Go to scopes folder to see [core scopes](scopes/core.md) and [extended scopes](scopes/extended.md) details.
```
ryanzhang-oss: What does "Application scope instances are defined and deployed by their standalone way" mean?

traits/core.md
ryanzhang-oss: The schematic is listed above. It seems that alone should suffice.

As for the Schema, do we have to force its format? It seems that we just gave two schemas format below.

scopes/core.md
negz: I'm not sure OAM should be defining the schema of these scopes at this point.

I know this was already part of the spec and thus changing it could be out of scope (hurr) for this PR, but I fear that this schema could be a little speculative and naive. Do we know of any implementations that use scopes in the wild? I recall Rudr does not support them, and we've found representing concepts like VPC networks in a portable, abstract fashion to be a hard thing(tm) in Crossplane.

scopes/core.md
```diff
@@ -0,0 +1,176 @@
...
+          "type": "string",
+          "description": "The type of the gateway, options are 'public', 'nat'. Empty string means no gateway."
+        }
+      }
+    }
+```
+
+### Instance Example
+
+For example, a network scope instance can be like below, asuming the version is `v1alpha1`:
```
negz:
```suggestion
For example, a network scope instance can be like below, assuming the version is `v1alpha1`:
```

scopes/core.md
negz: Not for scopes; they're distinct instances that are referenced by an `ApplicationConfiguration`. Traits are the only thing that gets embedded directly in an `ApplicationConfiguration`.

traits/core.md
```diff
@@ -0,0 +1,87 @@
+# Core traits
+
+## Manual Scaler
+
+This trait definition allows operators to manually scale the number of replicas for workload types that allow scaling operations.
+
+### Applicable Workloads
+
+ - `core.oam.dev/v1alpha2.containerWorkload`
```
negz: Can we agree on `ContainerisedWorkload` (or `ContainerizedWorkload` if we prefer en-us 😉)? In particular I'm pretty sure the lowercase `c` in `container` makes it an invalid CR kind.

negz: This `group/version.kind` representation of GVK still feels a little unusual to me - like we're using Kubernetes concepts (GVK) but inventing an OAM specific way of representing them. I wonder whether we should just stick to API convention and use separate fields. i.e.:

```yaml
spec:
  appliesTo:
  - apiVersion: core.oam.dev/v1alpha2
    kind: ContainerisedWorkload
```

One interesting side effect of this approach is that we could make `kind` optional, allowing a trait to apply to all core workload kinds by writing:

```yaml
spec:
  appliesTo:
  - apiversion: core.oam.dev/v1alpha2
```

traits/core.md
negz: Would it be worth picking a particular type of schema definition for all examples/reference schemas in the spec? Given we describe many of our non-schema examples using Kubernetes-compatible YAML I suggest we define schema using the OpenAPIv3 YAML syntax used below.

6.traits.md
ryanzhang-oss: > don't feel this would affect the spec or implementation that much, i.e.:

In this case, I wonder how many core workload Definitions are we going to put in the spec?

ryanzhang-oss: btw, the demo only uses Kind: server

traits/core.md
ryanzhang-oss: yes, we can use ContainerizedWorkload (we are in US 🥇 ).

scopes/extended.md
wonderflow: OK, I agree and I will remove it

6.traits.md
```diff
@@ -62,104 +62,26 @@ The top-level attributes of a trait define its metadata, version, kind, and spec
...
+The specification defines two major things: The list of workloads to which this trait applies, and the definition reference which will define the schema of this trait.
 
 | Attribute | Type | Required | Default Value | Description |
 |-----------|------|----------|---------------|-------------|
-| `appliesTo` | `[]string` | Y | `["*"]` | The list of workload types that this trait applies to. `"*"` means _any workload type_. A trait must apply to at least one workload type. This attribute must contain at least one value. An empty array is invalid for this attribute. |
-| `properties` | [`Properties`](#properties) | Y | | The trait's configuration options. The value is a [JSON schema](https://json-schema.org/) expressed as json string |
+| `appliesToWorkloads` | `[]string` | Y | `["*"]` | The list of workloads that this trait applies to. `"*"` means _any workload type_. A trait must apply to at least one workload type. This attribute must contain at least one value. An empty array is invalid for this attribute. |
+| `definitionRef` | [`DefinitionRef`](#definitionRef) | Y | | Index of the trait definition. |
```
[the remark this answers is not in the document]
wonderflow: Yeah, I think it's better.

scopes/core.md
```diff
@@ -0,0 +1,176 @@
+# Core scope type definitions
+
+The following core scope types are available:
+
+- network scope
+- health scope
+
+## Network scope
+
+The network scope groups components together and links them to a network or SDN. The network itself must be defined and operated by the infrastructure.
```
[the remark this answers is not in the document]
wonderflow: I think we mean `Application Operator` here.

traits/core.md
wonderflow: Maybe we could discuss about this way:

```
spec:
  appliesTo:
  - apiVersion: core.oam.dev/v1alpha2
    kind: ContainerizedWorkload
```

Currently, I'll change to 

```
  appliesToWorkloads:
    - core.oam.dev/v1alpha2.ContainerizedWorkload
```

6.traits.md
wonderflow: good idea.

traits/core.md
wonderflow: I'm not sure about this. In my view, we have to introduce K8s CRD if we want to finish the example of workflow for end user. Or we won't be able to tell, how we could get the schema, and how to find the version, and how the schema definition is really defined.

5.application_scopes.md
```diff
@@ -65,189 +46,21 @@ The spec defines the constituent parts of a scope.
...
```
ryanzhang-oss: Do we still have extended scopes definition?

6.traits.md
```diff
@@ -31,19 +31,11 @@ Any trait that is made available in a runtime MUST be implemented for that runti
...
-
- - **Extension traits.** The specification allows a runtime to define its own set of trait definitions that are unique to that runtime in addition to those defined by core and standard traits. A runtime can choose to implement any set of traits in this category with any definition. Extension trait type names MUST NOT be in the `core.oam.dev` or `standard.oam.dev` groups.
+There are some categories of traits, refer to [trait categories documentation](traits/categories.md) for more details.
```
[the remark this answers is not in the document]
ryanzhang-oss: Same question, why not make it the same as workloads categories section?

scopes/categories.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Extended application scope types
+
+Extended application scopes are optional per runtime, meaning that each runtime may choose which extended scopes are supported. It is expected that runtimes will implement extended scopes as they need. In this version of the spec, allowing user-defined extended application scope types is not supported.
```
ryanzhang-oss: The workload PR keeps those in the main spec while just move the definitions of them out. I wonder if that makes sense?

traits/core.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Schematic
+
+| Attribute | Type | Required | Default Value | Description |
+|-----------|------|----------|---------------|-------------|
+| `replicaCount` | `integer` | Y | `0` | The target number of replicas to create for a given component instance. |
+
+### Spec
+
+As a core trait, we don't need any trait definition here. But we could give this as an example.
```
ryanzhang-oss: what does this mean?

5.application_scopes.md
```diff
@@ -65,185 +65,18 @@ The spec defines the constituent parts of a scope.
```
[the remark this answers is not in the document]
zhxu2: Will we still be having health/network scopes as core scopes? Or will scopes be selfimplemented CRD concepts(extensibles)?

traits/core.md
hongchaodeng: @ryanzhang-oss 
Talked with Jianbo offline. I will help fix the grammars and expression. Will push to his branch later.

wonderflow:
I have rebased and updated the whole PR, please review it again. Thanks! ping @hongchaodeng @negz @hasheddan @vturecek @ryanzhang-oss

5.application_scopes.md
```diff
@@ -66,185 +59,29 @@ The spec defines the constituent parts of a scope.
...
-| `type` | `string` | Y | | Determines type of the application scope. The core types are defined [above](#core-application-scope-types) |
 | `allowComponentOverlap` | `bool` | Y | | Determines whether a component is allowed to be in multiple instances of this scope type simultaneously. When false, the runtime implementation MUST produce an error and stop deployment if an attempt is made to place a component into more than one instance of this scope type simultaneously. |
-| `parameters` | [`[]Parameter`](#parameter) | N | | The scope's configuration options. |
+| `definitionRef` | [`DefinitionRef`](#definitionRef) | Y | | Index of the scope definition. |
 
-The `type` field defines the type name for this type. Examples may be found in the [Core Application Scope Types](#core-application-scope-types) above. Extended scope types MUST NOT be in the `core.oam.dev` namespace.  However, they must follow the [group, version, kind notation](2.overview_and_terminology.md). For example: `scopes.example.com/v1.ExtendedNetworkScope`
+### DefinitionRef
 
-#### Parameter
+We use the same [DefinitionRef](3.workload.md#DefinitionRef) in workload.
```
hongchaodeng:
```suggestion
We use the same [DefinitionRef](3.workload.md#DefinitionRef) in _WorkloadDefinition_.
```

5.application_scopes.md
```diff
@@ -19,44 +20,36 @@ Components A, B, and C are deployed to the same health scope. The health scope w
...
- - Instances of core *workload* types MUST be deployed into an instance of each core application scope type. 
- - Runtimes MUST provide a default "root" application scope instance for each core application scope type.
- - Runtimes MUST deploy each component instances into default "root" application scope instances when an application configuration does not otherwise specify an application scope for a component instance.
+There are two kinds of application scope types:
 
- The following core application scope types are defined by this specification:
+ - **Core application scope types**: Core application scope types define grouping constructs for basic runtime behavior. They have the following characteristics:
+    - Core application scope types MUST be in the `core.oam.dev` namespace.
+    - Core application scope types MUST all be supported by any implementation of the specification. They may be implemented by any available feature in the platform, but they MUST be implemented according to the specification of each core application scope type.
+    - Instances of core *workload* types MUST be deployed into an instance of each core application scope type.
```
zhxu2: By "Instances of core *workload* types" are we referring to the workload CRs? Or are we referring to components that are configured to a certain kind of workload?

5.application_scopes.md
```diff
@@ -19,44 +20,36 @@ Components A, B, and C are deployed to the same health scope. The health scope w
...
 
-### Extended application scope types
+ - **Extended application scope types**: Extended application scopes are optional per runtime, meaning that each runtime may choose which extended scopes are supported. It is expected that runtimes will implement extended scopes as they need. In this version of the spec, allowing user-defined extended application scope types is not supported.
```
ryanzhang-oss:
```suggestion
 - **Extended application scope types**:  Each runtime MAY choose to implement any number of extended application scopes. 
```

5.application_scopes.md
wonderflow: It means workload CRs, I should fix it

5.application_scopes.md
```diff
@@ -66,185 +59,29 @@ The spec defines the constituent parts of a scope.
...
+| `definitionRef` | [`DefinitionRef`](#definitionRef) | Y | | Index of the scope definition. |
 
-The `type` field defines the type name for this type. Examples may be found in the [Core Application Scope Types](#core-application-scope-types) above. Extended scope types MUST NOT be in the `core.oam.dev` namespace.  However, they must follow the [group, version, kind notation](2.overview_and_terminology.md). For example: `scopes.example.com/v1.ExtendedNetworkScope`
+### DefinitionRef
 
-#### Parameter
+We use the same [DefinitionRef](3.workload.md#DefinitionRef) in _WorkloadDefinition_.
 
-The parameters that a scope exposes to operators.
+## Core application scopes
```
resouer: I suggest we don't define core application scopes for now? Either Network Scope or Health Scope does not sound like a MUST implement to me. Maybe move them to standard scope @ryanzhang-oss

core/traits/manual_scaler_trait.md
```diff
@@ -0,0 +1,89 @@
...
+metadata:
+  name: manualscalertrait.core.oam.dev
+spec:
+  appliesToWorkloads:
+    - core.oam.dev/v1alpha2.ContainerizedWorkload
+  definitionRef:
+    name: manualscalertrait.core.oam.dev
+```
+
+Then OAM runtime should have a version of this schema, the schema SHOULD be able to validate. For example using JSON Schema like this:
```
resouer: Maybe add a CRD as example as well?

core/traits/manual_scaler_trait.md
```diff
@@ -0,0 +1,89 @@
...
+```yaml
+apiVersion: core.oam.dev/v1alpha2
+kind: ApplicationConfiguration
+metadata:
+  name: custom-single-app
+  annotations:
+    version: v1.0.0
+    description: "Customized version of single-app"
+spec:
+  variables:
```
resouer: delete?

core/traits/manual_scaler_trait.md
negz: Perhaps a CRD-compatible OpenAPIv3 schema? This would make it easy to relate the schema to a Kubernetes CRD without leaking concrete Kubernetes specifics into the spec.

5.application_scopes.md
wonderflow: totally agree

core/traits/manual_scaler_trait.md
wonderflow: OK I'll add a CRD-compatible OpenAPIv3 schema

core/traits/manual_scaler_trait.md
wonderflow: sure

5.application_scopes.md
zhxu2: lgtm

5.application_scopes.md
zhxu2: agree

PR #321, opened 2020-02-29
1.0.0-alpha2 changes

hongchaodeng:


2.overview_and_terminology.md
```diff
@@ -20,38 +20,57 @@ This specification proposes a model that defines cloud native applications as fo
...
+- __Application configuration__, defines a deployment of components, their traits, and application scopes.
 
-Thus, an _application_ is a collection of _components_ with a set of operational traits and scoped together into one or more application boundaries. 
+In addition, the specification defines the following definition constructs for infrastructure operators and platform builders to define the components, traits, and scopes that are available to application developers and application operators on a given platform:
+
+- __Workload Definition__, where platforms declare the schema of a specific kind of workload, such as containers, functions, or data stores, that developers use to create components.
+- __Trait Definition__, where platforms declare the schema of a specific kind of trait.
+- __Scope Definition__, where platforms declare the schema of a specific kind of application scope.
+
+Thus, OAM defines an _application_ as a collection of _components_ each with a set of operational traits and scoped together into one or more application boundaries.
```
artursouza: Can we have a diagram here? See the one I created below, it might help:

image

core/traits/manual_scaler_trait.md
```diff
@@ -0,0 +1,125 @@
...
+    - componentName: frontend
+      traits:
+        - trait:
+            apiVersion: core.oam.dev/v1alpha2
+            kind: ManualScalerTrait
+            spec:
+              replicaCount: 5
+```
+
+This example illustrates using the manual scaler trait to manually scale a component by specifying the number of replicas the runtime should create. This trait has only one attribute in its properties: `replicaCount`. This takes an integer, and a value is required for this trait to be successfully applied. If, for example, an application configuration used this trait but did not provide the `replicaCount`, the system would reject the application configuration.
```
artursouza: - This takes an integer, and a value is required for this trait to be successfully applied.
+ This is a required property that takes a positive integer.

7.application_configuration.md
```diff
@@ -0,0 +1,186 @@
...
+          spec:
+            replicaCount: 3
+      scopes:
+        - scopeRef:
+            apiVersion: core.oam.dev/v1alpha2
+            kind: NetworkScope
+            name: example-vpc-network
+```
+
+The example above illustrates a complete application configuration scopes, component instances, their traits, and parameter overrides that allow users to specify values at deployment time from the command line.
```
artursouza: - The example above illustrates a complete application configuration scopes, component instances, their trait, and ...
+ The example above illustrates a complete application configuration, containing scopes, component instances with corresponding traits, and ...

2.overview_and_terminology.md
negz: I think a diagram is a good idea. The suggested diagram mostly makes sense, but is not completely accurate. I would say:

Definitions:

* A trait definition defines a kind of trait - this is accurate in the diagram.
* A workload definition defines a kind of workload. A `kind: WorkloadDefinition` defines `kind: ContainerizedWorkload`.
* A scope definition defines a kind of scope. A `kind: ScopeDefinition` defines `kind: NetworkScope`.

Types

* An application scope is not a type of scope definition. A scope definition defines a type of application scope.
* It may be clearer to think of a component as _templating_ a type of workload. A component is not a type of workload.

Implementations

* `NetworkScope`, `ContainerizedWorkload`, and `ManualScalerTrait` implement scope, workload, and trait (which are concepts, or duck types). Not scope definition, workload definition, and trait definition.

artursouza: image

ryanzhang-oss: Thanks, Artur. While a picture worth a thousand words, I am having some problems understanding this diagram.  Just to name a few

- I don't know what does the right side (yellowish part) of the graph represents
- I am not sure if I will say appConfig "config" a scope or a trait
- I am also not sure if a component is  "a type of " workloadDefinition. Maybe it is a type of workload?
- I am also not sure if a containerizedWorkload "implements" a workloadDefinition.

hongchaodeng:
@negz @artursouza 

Would you mind to send PRs to`1.0.0-alpha2` branch?
Then this PR will get updated automatically.

2.overview_and_terminology.md
artursouza: Thanks for your feedback. Let me answer your questions:

1. Yellow, means it is an implementation of OAM and not OAM spec itself.
2. It seems that "config" was the most natural verb for relationships for "AppConfig" - any ideas of which verbs would be more appropriate? Maybe "injects Traits" and "defines Scopes" ?
3. Diagram has been updated since that comment. Please, check latest version above.
4. Same. Diagram is updated.

wonderflow:
I think everything is almost finished. Can we merge this to master?

hongchaodeng:
In today’s meeting, we have agreed on moving forward this major alpha2 spec changes and fixed the minor issues later.

@vturecek @resouer Would you take a final look and approve it?

hongchaodeng:
Thanks @vturecek !
Let's merge this and move forward the alpha2 spec.

PR #322, opened 2020-03-01
[1.0.0-alpha2]Rewording and example fixes on Overview and Workload

artursouza:
* Example for same name, same kind but different group.
* Fix one example in WorkloadDefinition
* Rewording to match RFC 2119
* Ad-hoc rewording

### DIFF
--- a/2.overview_and_terminology.md
+++ b/2.overview_and_terminology.md
@@ -145,6 +145,19 @@ metadata:
     name: foo
 ```
 
+*Okay:* Each group (`one.dev` and `other.dev`) allows the kind `Component` with name `foo`.
+```yaml
+apiVersion: one.dev/v1alpha2
+kind: Component
+metadata:
+    name: foo
+---
+apiVersion: other.dev/v1alpha2
+kind: Component
+metadata:
+    name: foo
+```
+
 *NOT Okay:* Version is not a namespace qualifier.
 
 ```yaml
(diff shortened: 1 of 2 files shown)


2.overview_and_terminology.md
```diff
@@ -145,6 +145,19 @@ metadata:
     name: foo
 ```
 
+*Okay:* Each group (`one.dev` and `other.dev`) allows the kind `Component` with name `foo`.
```
hongchaodeng: What's the motivation to add this paragraph?

2.overview_and_terminology.md
```diff
@@ -176,7 +189,7 @@ Annotations provide a mechanism for attaching arbitrary text within the metadata
 
 > Annotations are key/value pairs. Valid annotation keys have two segments: an optional prefix and name, separated by a slash (/). The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between. The prefix is optional. If specified, the prefix must be a DNS subdomain: a series of DNS labels separated by dots (.), not longer than 253 characters in total, followed by a slash (/).
 
-The following annotation labels are _predefined_ and are _strongly recommended_.
+The following annotation labels are _predefined_ and NOT RECOMMENDED.
```
ryanzhang-oss: why is this not recommended with an example using it right below?

2.overview_and_terminology.md
artursouza: It was not clear that a different group is enough. It is mentioned in the text but without example.

2.overview_and_terminology.md
artursouza: My bad. The “NOT” is not supposed to be there.

artursouza: Fixed

hongchaodeng:
LGTM

Issue #23, opened 2019-07-03
component startup ordering


xiang90:
The components in the operationalConfiguration is a flat array right now. In some cases, we might want to express the dependencies of components. Component A might need to start before Component B, or Component B might fail. 

We might solve this problem by blindly retry in starting phase of components, but ideally we can express ordering requirement explicitly.

Issue #171, opened 2019-09-27
Dependency story and its solutions


hongchaodeng:
As mentioned before, I have been collecting **dependency** use cases from our users and trying to summarize them in an organized and clarified manner.

Here's the final work of my study: [Dependency Story in OAM](https://docs.google.com/document/d/1TSr6eHWz5g-uA5-M4e9MKty_Lvuyl0liS7YihFjKnPY/edit?usp=sharing)

The doc provides categorization and solution for our problems. Feel free to comment, give feedback. Eventually, we want to add such tutorials in the repo and enable users to follow with examples.

Issue #323, opened 2020-03-03
Reconciling OAM traits in v1alpha2


negz:
#### Background

My understanding is that currently in the Rudr implementation of the v1alpha1 OAM spec all knowledge of how to render an `ApplicationConfiguration` into into Kubernetes specific kinds (e.g. `Deployment` and `Service`) is centralised in one controller (or rather [instigator](https://github.com/oam-dev/rudr/blob/d674dc50d24ee18125d007e33b2ada3a5e23eeaf/src/instigator.rs)). I don't speak Rust, but it appears that this controller:

1. Watches for `ApplicationConfiguration`
1. Gets each `ComponentSchematic` referenced by the `ApplicationConfiguration`
1. Creates, updates, or deletes Kubernetes resources as appropriate to satisfy the workload described by the `ComponentSchematic`. For example by rendering and creating a `Deployment`, `Service`, etc.
1. Creates, updates, or supplements the Kubernetes resources produced in step 3 per the traits associated with the component that produced them.

This allows a lot of flexibility because one controller is authoritative for everything to do with an `ApplicationConfiguration`, but could impact the maintainability and extensibility of the code. The `exec` method of the instigator is ~300 lines long, and support for new kinds of workloads and traits must be compiled into the instigator.

In the [nascent Crossplane implementation](https://github.com/crossplane/crossplane/pull/1321) of the v1alpha2 OAM spec an `ApplicationConfiguration` is reconciled by _a series of controllers_.  The `ApplicationConfiguration` controller:

1.  Watches for `ApplicationConfiguration`.
1. Gets each `Component` referenced by the `ApplicationConfiguration
1. Renders the `Component` into a Kubernetes CR representing its workload kind, e.g. `ContainerizedWorkload`.

At this point the job of reconciling the OAM workload is passed onto a workload-specific controller. For example a `ContainerizedWorkload` controller might:

1. Watch for `ContainerizedWorkload`
1. Create a `Deployment` and a `Service` for each `ContainerizedWorkload`

Note that the `ApplicationConfiguration` controller is, in a way, completely workload-unaware. It simply renders out a templated workload and defers the task of figuring out how to reconcile that workload into another controller.

[...]

PR #326, opened 2020-03-05
Resource Dependency in OAM

hongchaodeng:
In https://github.com/oam-dev/spec/issues/171, we have discussed our dependency stories and started to explore solutions.
Now that we have designed and implemented mechanism to satisfy dependency requirements, we are proposing to add them into OAM spec.

Fixes https://github.com/oam-dev/spec/issues/23


hongchaodeng:
cc @vturecek @negz @artursouza

design/20200305-spec-v1alpha2-dependency-parampassing.md
```diff
@@ -0,0 +1,94 @@
...
+### Parameter Passing
+
+For Case 2, while the above mechanism can provide the dependency ordering, it doesn't fulfill the parameter passing requirement. As a result, we propose to extend _ParameterValue_ to accept value from another component's field:
+
+
+```yaml
+kind: ApplicationConfiguration
+spec:
+  components:
+  - componentName: nas-file-system
```
wonderflow: We'd better add a note for `nas-file-system`:
The resource created by `nas-file-system` will output a `filesystemID` which finally aggregated at `.status`

artursouza:
LGTM.

design/20200305-spec-v1alpha2-dependency-parampassing.md
```diff
@@ -0,0 +1,96 @@
...
+2. MySQL connection info needs to be injected into Web's environment variables or files;
+
+### Case 2: NAS FileSystem and MountTarget
+
+Two components, NAS_FileSystem and NAS_MountTarget exists in the same AppConfig and NAS_MountTarget depends on NAS_FileSystem that:
+
+1. NAS_FileSystem needs to be created first;
+2. NAS_MountTarget needs filesystemID from NAS_FileSystem to fill in NAS_MountTarget's spec.
+
+## Proposal
```
ryanzhang-oss: Can you explain a bit on why these two cases are different?

design/20200305-spec-v1alpha2-dependency-parampassing.md
```diff
@@ -0,0 +1,96 @@
...
+- components:
+  - componentName: mysql
+  - componentName: web
+    dependencies:
+      - mysql
+```
+
+Here a component can describe its dependencies, or dependent components.
+
+For Case 1, we also suggest to use ServiceBinding in [trait-injector](https://github.com/oam-dev/trait-injector) to inject secret. It is a trait that should be applied before component starts. We propose to add `ordering` to express trait's ordering w.r.t the component.
```
ryanzhang-oss: I wonder if the ServiceBinding trait should be part of OAM spec? If so, should it be a core or a standard trait?

design/20200305-spec-v1alpha2-dependency-parampassing.md
```diff
@@ -0,0 +1,96 @@
...
+        spec:
+          bindings:
+          - from:
+              secret:
+                name: mysql-conn
+            to: # inject into Web Deployment's file path
+              filePath: /input/mysql-conn
+```
+
+### Parameter Passing
```
ryanzhang-oss: Is there any reason why we can't use serviceBinding trait here too?

design/20200305-spec-v1alpha2-dependency-parampassing.md
wonderflow: Let me explain more.

The first case represents different OAM implementation cooperate with each other. In other word, system of Web component don't know what will output from system of Mysql component. 

The second case represents two components are in the same system, they can passthrough arguments with each other. One can output a key and finally fill in to the other one's spec.

design/20200305-spec-v1alpha2-dependency-parampassing.md
wonderflow: I think it should be a core trait. ServiceBinding will no doubt be one of the shining point of OAM.

design/20200305-spec-v1alpha2-dependency-parampassing.md
wonderflow: I have explained why these two cases are different. So parameter passing here is more likely to be used as internal arguments delivery, while service binding aims to be used in a more common way.

wonderflow: More than that, we don't want to make service binding be coupled with parameters. Parameters are invented for application developers to tell what they need to application operators, while app operator can override the parameter. We could take parameter passing as an advanced override, it allows app operator override parameter from component output. 

In this view, they have nothing to do with service binding, right?

design/20200305-spec-v1alpha2-dependency-parampassing.md
```diff
@@ -0,0 +1,96 @@
...
+  dependencies:
+    - mysql
+  traits:
+    - ordering: preStart # applied before component starts
+      trait:
+        kind: ServiceBinding
+        spec:
+          bindings:
+          - from:
+              secret:
```
negz: Is this referring to a Kubernetes `kind: Secret`, or the generic platform-agnostic concept of named secret data?

design/20200305-spec-v1alpha2-dependency-parampassing.md
negz: If I follow correctly, the problem we're solving here is that when workload A is created by appconfig X it may want to depend on some property of workload B, which is also created by appconfig X. It's possible that the property workload A depends on will be sensitive (e.g. a shared secret).

It seems like we're proposing two ways of addressing this:

1. The `ServiceBinding` trait, which requests that the data from a named secret (which is presumably produced by worklad B) be mounted at a named path within workload A.
1. Parameter passing, which requests that a parameter of workload A be set to the value of a fieldpath within workload B.

It seems like parameter passing is simpler, but `ServiceBinding` is more flexible. Would OAM be simpler overall if we supported only `ServiceBinding` and not parameter passing?

wonderflow: `parameter passing` is also required as it gives us a way to pass through parameters directly between components. By using `ServiceBinding` trait, we are able to pass through `object` level information between components, while `parameter passing` can support `spec field` level information.

design/20200305-spec-v1alpha2-dependency-parampassing.md
wonderflow: I think it's generic platform-agnostic concept of named secret data for OAM spec. And in OAM K8s implementation it's `kind:Secret`

design/20200305-spec-v1alpha2-dependency-parampassing.md
```diff
@@ -0,0 +1,96 @@
...
+In https://github.com/oam-dev/spec/issues/171, we have discussed our dependency stories and started to explore solutions.
+Now that we have designed and implemented mechanism to satisfy dependency requirements, we are proposing to add them into OAM spec.
+
+## Use Cases
+
+### Case 1: MySQL and Web
+
+Two components, MySQL and Web exists in the same AppConfig and Web depends on MySQL that:
+
+1. MySQL needs to start first;
```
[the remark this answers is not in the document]
wonderflow: It requires the consumer has the ability to recovery. But we do have cases that user's workload which can't self-healing. So in the view of platform builder, we could add a little more into our application model.

design/20200305-spec-v1alpha2-dependency-parampassing.md
negz: Could you help me understand what the difference between object level and spec field level information?

wonderflow: For example: 

Assuming below is an existing CRD + Operator which can receive filesystemID as part of spec in CR.

```
   apiVersion: ros.aliyun.com/v1alpha2
   kind: NASMountTarget
   spec:
      filesystemID: "" # to be filled
```
If we give a secret object to it, it has no way to consume this secret.

While if the workload is K8s deployment, it has the way to consume secret object:

```
apiVersion: apps/v1
kind: Deployment
spec:
  template:
    spec:
      containers:
        envFrom:
        - secretRef:
            name: my-secret
```

design/20200305-spec-v1alpha2-dependency-parampassing.md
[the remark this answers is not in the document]
hongchaodeng: Hi @negz. We have the same conservation even back to the beginning: https://github.com/oam-dev/spec/issues/23.

Initially, we want to rely on component itself to restart or hold back. But there are many existing scenarios that requires this dependency functionality for their apps to work properly. For example, we have OAM users whose apps can't start and trigger alerts. I think this feature is a major step to migrate as many existing apps as we could.

design/20200305-spec-v1alpha2-dependency-parampassing.md
```diff
@@ -0,0 +1,96 @@
...
+Here a component can describe its dependencies, or dependent components.
+
+For Case 1, we also suggest to use ServiceBinding in [trait-injector](https://github.com/oam-dev/trait-injector) to inject secret. It is a trait that should be applied before component starts. We propose to add `ordering` to express trait's ordering w.r.t the component.
+
+```yaml
+- componentName: web
+  dependencies:
+    - mysql
+  traits:
+    - ordering: preStart # applied before component starts
```
artursouza: Having dependency information makes sense. On the other hand, specifying at which stage something should be applied is imperative and not declarative. IMO, OAM should keep a declarative config as much as possible.

hongchaodeng: Could you provide a suggestion for declarative approach?

artursouza: I was thinking on a way to infer when applying the config that it needs to run in preStart phase. Is there a scenario binding will happen in another phase? If not, we can simply automatically apply bindings on preStart - always. Then, the `ordering` attribute can be removed from the spec.

design/20200305-spec-v1alpha2-dependency-parampassing.md
```diff
@@ -0,0 +1,96 @@
...
+      trait:
```
hongchaodeng: trait ordering defined in TraitDefinition?

ryanzhang-oss: I like that idea but I wonder if there are any cases that a trait can be both pre and post component?

hongchaodeng:
Summarize the discussion in today's meeting:

- _Dependency_ needs details in user stories and how it simplifies the app development. @vturecek also mentioned app-level fault tolerance which we had discussed in https://github.com/oam-dev/spec/issues/23. The use cases we have seen, including community ones, have requirements on dependency due to no app-level fault tolerance. I am going to add them in the design doc.

- _Trait Ordering_ could be put into TraitDefinition if there is no use case that the same kind of Trait could be both pre and post component.

- _Field Passing_ has no objection afaik.

- _Service Binding_ is one way that we could define _source data_ and _target workload_ and bind them automatically. Since it is not part of the design we might separate it out into another doc.

negz:
> Trait Ordering could be put into TraitDefinition if there is no use case that the same kind of Trait could be both pre and post component.

I believe it would also be useful for this design to consider and discuss how a runtime would actually enforce/enact trait ordering. I think this has overlap with the topics we've discussed in https://github.com/oam-dev/spec/issues/323. What does a component "starting" mean? How do we know when the component is started? How do we prevent the component starting until the trait is applied?

negz:
@hongchaodeng I wonder if this could be an alternative to the `ServiceBinding` concept:

```yaml
---
# The Component that represents the server process.
kind: Component
metadata:
  name: coolcw
spec:
  workload:
    kind: ContainerizedWorkload
    spec:
      containers:
      - name: coolcontainer
        image: cool/container:latest
        # The app developer declares that they expect the connection details for
        # MySQL to exist at filesystem path /input/mysql-conn. They do not need
        config:
        - path: /input/mysql-conn
          # We allow the OAM spec to specify that config files (and env vars)
          # can be populated from a 'Secret'. This does not mean OAM is aware of
          # Kubernetes kind: Secret, but it does mean OAM runtimes must support
          # some concept like it. We can omit fromSecret here because it will be
          # set by a parameter.
          #
          # fromSecret: mysql-secret
  parameters:
  - name: instance-name
    description: The name of the containerized workload.
    required: true
    fieldPaths:
    - "metadata.name"
  # The app developer allows the app operator to specify the name of the secret
  # that will be mounted inside the container at /input/mysql-conn.
  - name:  mysql-secret
    description: |
      The secret from which the containerized workload should read its
      connection details, such as username, password, and endpoint.
    required: true
    fieldPaths:
    - "spec.containers[0].config[0].fromSecret"
---
# The Component that represents the MySQL database (i.e the default database of
# the requested MySQL instance).
kind: Component
metadata:
  name: coolsql
spec:
  workload:
    apiVersion: database.crossplane.io/v1alpha1
    kind: MySQLInstance
    spec:
      engineVersion: 5.6
      # Crossplane managed resources typically generate and write credentials,
      # and other connection details (e.g. endpoints) rather than reading them.
      # This configures which secret Crossplane should write the connection
      # details for this MySQL instance to. We can omit it here because it will
      # be set by a parameter.
      #
      # writeConnectionSecretToRef:
      #   name: mysql-secret
  parameters:
  - name: instance-name
    description: The name of the MySQL instance.
    required: true
    fieldPaths:
    - "metadata.name"
  # The app developer allows the app operator to specify the name of the secret
  # that will be written by Crossplane when the MySQLInstance becomes available.
  - name:  mysql-secret
    description: |
      The secret to which the MySQL instance should write its connection
      details, such as username, password, and endpoint.
    required: true
    fieldPaths:
    - "spec.writeConnectionSecretToRef.name"
---
metadata:
  name: coolapp
spec:
  components:
  - componentName: coolsql
    parameterValues:
      - name: instance-name
        value: coolapp-sql
      # The app operator specifies where to write the secret...
      - name: mysql-secret
        value: coolsecret
  - componentName: coolcw
    parameterValues:
      - name: instance-name
        value: coolapp-server
      # ...and where to read the secret.
      - name: mysql-secret
        value: coolsecret
```

This seems like a less complicated way to support passing sensitive connection details between infrastructure and applications than the `ServiceBinding` proposal. It shifts the responsibility of injecting the secret config file (or env vars) to the `ContainerizedWorkload` implementation, rather than a trait. I think this fits well with the OAM separation of concerns, since it will typically be the developer (who writes the Component) asserting that "my application expects connection details to exist at this path, and those details are secret".

For Kubernetes runtimes it should be quite easy to translate this to the containers array of a `podTemplateSpec`, and rely on the kubelet's built in support for [mounting and exposing Kubernetes secrets inside a pod](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod) as env vars or config files.

hongchaodeng:
We  have discussed about trait ordering within our platform teams and have agreement that trait ordering should be in TraitDefinition because:

- A trait provider would know the details whether it should be before or after a workload.
- Initially, we can add a simple `stage` field in TraitDefinition to apply to all workloads. If we found that we have different ordering requirements for different workloads, we can add that to TraitDefinition later.

cc @vturecek @negz

hongchaodeng:
@negz 
What a wonderful idea! I actually like that a lot and think it could be in core spec.

Nonetheless, there is still one pitfall for us. First of all, let me diff on your idea (abbr. new idea) and the current proposal:

1. The new idea discusses how to write secret name, which current proposal doesn't handle. We can merge and I would like to use the new idea.

2. In the new idea, how to read the secret and mount it into path is handled in _Workload_, while current proposal handles it in _Trait_.

The impact of point 2 has impacted us in two ways:

1. We have a variety of extended container Workload which are different CRDs but all translate to Deployment. First of all we can't guarantee the extended Workload could have this config field since it is not core core Workload. Moreover, it is duplicate work for all of extended Workload to implement secret mounting. Thus we want Traits for these common logic.

2. The concept of secret is leaked into Components. Components are for developers, and some developers don't know or want to know "secret". It's also the reason that some extended Workload is different from core ContainerizedWorkload. Their developers might have different level of skills and some doesn't know about ports, volumes, health probes.. They just say run my app, give it a file.

But generally, I like the new idea and I think it is more natural to OAM and we could align on that in the core spec.

ryanzhang-oss:
> This seems like a less complicated way to support passing sensitive connection details between infrastructure and applications than the `ServiceBinding` proposal.

@negz I wonder if this also replaced the parameter passing method? If not, can you explain the differences so that I can understand the nuances? 

Another question is, as I eluded to in the meeting,  I wonder how can this apply to workloads that are maintained by a third party team that refused to change their schema to include such fields in their workload definition?

artursouza:
One concern from @negz's proposal is the readability of the reference inversion where fromSecret (the consuming end) is set at the resource declaration instead (param). I honestly found it hard to navigate. Same theme I provided feedback on params for v1alpha2.

Second, secret management seems to be a common use case that it deserves its own concept, not just another Trait defintion.

Dependency management is a separate problem than secret management. Dependency management is only applicable in the second scenario since the creation of the second resource might not be possible without the first. It means that OAM implementations must implement a workflow engine.

## SQL and Web

I would remove the explicit dependency and let the app itself handle dependency health. For secret, OAM spec would have explicit secret store attribute (which is a new concept called SecretStore) and keep the fromSecret for paramaterValues.

The SecretStores would be provided by the runtime implementation (just like Traits).

```  
metadata:
  name: service-tracker
spec:
  secretStore:
    apiVersion: extension.oam.dev/v1alpha2
    kind: AzureKeyvault
    name: azure-sql
    parameterValues:
      authFileMount: "/mnt/secure/myauth.txt"
  components:
    - componentName: tracker-postgres-db
      parameterValues:
        - name: dbuser
          fromSecret: mysecretuser
        - name: dbpassword
          fromSecret: mysecretpassword
        - name: database
          value: hackfest  
    - componentName: data-api
      instanceName: data-api
      parameterValues:
        - name: dbuser
          fromSecret: mysecretuser
        - name: dbpassword
          fromSecret: mysecretpassword
        - name: dbhostname
          value: "postgres.default.svc.cluster.local"
        - name: dbport
          value: "5432"
        - name: dbname
          value: "hackfest"  
        - name: dbdriver
          value: "postgres"
        - name: dboptions
          value: ""

```

If fieldPaths is a must (to keep consistency), then the secretStore would declare each secret to be consumed and where it should be populated. This would be the same as above but with fieldPaths:

```  
metadata:
  name: service-tracker
spec:
  secretStore:
    apiVersion: extension.oam.dev/v1alpha2
    kind: AzureKeyvault
    name: azure-sql
    parameterValues:
      authFileMount: "/mnt/secure/myauth.txt"
   secrets:
     - secret:
        name: mysecretuser
        fieldPaths:
          - "specs.components[0].paramValues.[?(@.name=='dbuser')].value"
          - "specs.components[1].paramValues.[?(@.name=='dbuser')].value"
     - secret:
        name: mysecretpassword
        fieldPaths:
          - "specs.components[0].paramValues.[?(@.name=='dbpassword')].value"
          - "specs.components[1].paramValues.[?(@.name=='dbpassword')].value"
  components:
    - componentName: tracker-postgres-db
      parameterValues:
        - name: dbuser
          value:
        - name: dbpassword
          value:
        - name: database
          value: hackfest  
    - componentName: data-api
      instanceName: data-api
      parameterValues:
        - name: dbuser
          value:
        - name: dbpassword
          value:
        - name: dbhostname
          value: "postgres.default.svc.cluster.local"
        - name: dbport
          value: "5432"
        - name: dbname
          value: "hackfest"  
        - name: dbdriver
          value: "postgres"
        - name: dboptions
          value: ""

```

negz:
> I wonder if this also replaced the parameter passing method? If not, can you explain the differences so that I can understand the nuances?

@ryanzhang-oss In my proposal we presume workloads communicate the values they need to share via some kind of secret store in which a secret can be uniquely identified by name - possibly just a Kubernetes `Secret`. This implies workload B could only depend on settings that workload A wrote to the secret store.

The parameter passing proposal allows workload B to depend on any API field of workload A. This is more flexible because it allows us to reference any field, even those that are not written to the secret store. However parameter passing cannot help with sensitive data, which we don't want to expose as a workload field (e.g. under `status`). 

All in all, I like the parameter passing API but am wary of the complexity it would bring to the AppConfig controller implementation. It seems like it would require either implicit or explicit dependency ordering (if workload B uses a field from workload A then workload B could not created until workload A was created). It also seems like it would expand the responsibility of the AppConfig controller to include reading back the workloads it creates in order to reference their fields.

hongchaodeng:
Updated the doc.

> I would remove the explicit dependency and let the app itself handle dependency health.

I would argue that this is inaccurate. Let me explain more with a real world example in the following:

Let's say I have a ML app that has three components:

- preprocess
- training
- serving

This is true for most ML frameworks and they have strict ordering for:

```
  preprocess -> training -> serving
```
Without the previous stage finished, the next could not start due to lack of data.
Today, ML framework including Tensorflow would not handle such dependency health. They rely on the platform to handle it.
**If we want to define ML apps in OAM, this is a must-have feature.**

negz:
> If we want to define ML apps in OAM, this is a must-have feature.

I'm still not convinced that the OAM application model has to handle these concepts. It seems like it would be possible for the workloads involved to model this via eventual consistency. Are there existing Kubernetes CRs that model ML flows we could look to for guidance?

wonderflow:
For dependency, Not only the ML app, but also all legacy apps which won't have self healing need this dependency feature. We can't say OAM won't support these apps unless they have self healing feature.

However, I think we could split this proposal into small pieces so we could discuss with more focus

hongchaodeng:
Referring a similar request in oam-dev mailing list: https://groups.google.com/forum/#!topic/oam-dev/X4npG1ay-sc

hongchaodeng:
>  It seems like it would be possible for the workloads involved to model this via eventual consistency. Are there existing Kubernetes CRs that model ML flows we could look to for guidance?

I could agree with you from technical perspective. But many real world apps I have seen require such feature.

Kubeflow is one example.

hongchaodeng:
> I think we could split this proposal into small pieces so we could discuss with more focus

@wonderflow Could you clarify what you mean by "small pieces"?

wonderflow:
> > I think we could split this proposal into small pieces so we could discuss with more focus
> 
> @wonderflow Could you clarify what you mean by "small pieces"?

I mean should we split this proposal into three pieces?

1. trait ordering: it seems no one object it?
2. field passing by parameter:  no one object it too.
3. component dependency: more discussion needed.

ryanzhang-oss:
> @wonderflow Could you clarify what you mean by "small pieces"?

As we discussed in the [last community meeting ](https://docs.google.com/document/d/1nqdFEyULekyksFHtFvgvFAYE-0AMHKoS3RMnaKsarjs/), we will split this proposal into at least the following two propsoals
- how to pass information between components
- how to define dependencies between components

ryanzhang-oss:
> It seems like it would be possible for the workloads involved to model this via eventual consistency.

I agree but isn't the entire v1alpha2 spec change done to allow existing applications to adopt OAM __WITHOUT__ having to change?

hongchaodeng:
Created an issue https://github.com/crossplane/oam-kubernetes-runtime/issues/11.

We will start with annotation-based implementation based on this design in crossplane.

design/20200305-spec-v1alpha2-dependency-parampassing.md
```diff
@@ -0,0 +1,102 @@
...
+  components:
+  - componentName: nas-file-system # The Workload of nas-file-system output a filesystemID at .status.filesystemID
+
+  - componentName: nas-mount-target
+    dependencies:
+    - nas-file-system
+    parameterValues:
+    - name: filesystem-id
+      from:
+        compoent: nas-file-system
```
draveness, line 86: nit: component

design/20200305-spec-v1alpha2-dependency-parampassing.md
```diff
@@ -0,0 +1,102 @@
...
+## Background
+
+In https://github.com/oam-dev/spec/issues/171, we have discussed our dependency stories and started to explore solutions.
+Now that we have implemented and evolved the design to solve real world problems, we are proposing the design to OAM spec.
+
+## Goals
+
+In this design, we want to help applications reduce the burden of handling fault tolerance and dependency health.
+By putting this work in the OAM platform layer, OAM users can define explicit dependency and parameter passing in a standard way
+without carign about the implementation details.
```
pigletfly, line 14: typo carign > caring

pigletfly:
It's a wonderful proposal.There is a scenario as below:

a [nsq](https://github.com/nsqio/nsq) cluster is composed with three components, that is nsqd nsqlookup and nsqadmin.
If we want to setup a nsq cluster, we should make sure nsqd is up before nsqlookup, then run a nsqadmin workload at last.

That is, nsqlookup consumes the provide of nsqd component instance, nsqadmin consumes the provide of nsqlookup component instance.

If we couldn't describe the dependencies order between the components, the component could run properly.

This is similarly to https://github.com/kubernetes/kubernetes/issues/65502 , the issue describes dependencies between containers on the same Pod, we are discussing dependencies between components.

hongchaodeng:
I have updated the proposal with latest design.
It's been more formalized with more implementation details and code examples.
I will start implementing this proposal in next couple of weeks in https://github.com/crossplane/oam-kubernetes-runtime

hongchaodeng:
Since it's no change on OAM spec and more implementation focused, we are moving the design into oam-kubernetes-runtime: https://github.com/crossplane/oam-kubernetes-runtime/pull/24

wonderflow:
We'd better keep the old proposal as a ref and make this new one implemented first. Then in `v1alpha3` version or `v1beta1`, we could make the old one happen. 

And we can provide an upgrade path if one day we will make dependency into spec instead of trait.

negz:
>Since it's no change on OAM spec and more implementation focused, we are moving the design into oam-kubernetes-runtime: crossplane/oam-kubernetes-runtime#24

@hongchaodeng Should we close this PR? I went to review the PR you opened against oam-kubernetes-runtime, but accidentally started writing all my comments this one.

hongchaodeng:
@negz Yeah. Let's converge the work to crossplane/oam-kubernetes-runtime#24

PR #327, opened 2020-03-08
[1.0.0-alpha2] Fix target audience to include all parties

hongchaodeng:


4.component.md
```diff
@@ -1,5 +1,5 @@
 # 4. The Component Model
-**Target Audience: Application Developer** 
+**Target Audience: Application Developer, Application Operator**
```
ryanzhang-oss: What does an application operator do with a component?

wonderflow: Mainly to know what parameters they can passthrough.

hongchaodeng:
> While I think the idea is right, it kinda makes the separation of concern not that "separate" anymore.

Let's make it right first and make it better together down the path.

PR #335, opened 2020-03-16
[1.0.0-alpha2] rectify parameter fieldPath to have leading dot

hongchaodeng:
I have walked through the spec over the weekends and found this discrepancy with the [design doc](https://github.com/oam-dev/spec/blob/master/design/20200105-spec-v1alpha2-kubernetes-friendly.md)

### DIFF
--- a/4.component.md
+++ b/4.component.md
@@ -49,7 +49,7 @@ The Parameters section defines all of the configurable parameters for this compo
 
 Parameter `name` fields must be Unicode letter and number characters. Application Configurations will specify parameters values using this name.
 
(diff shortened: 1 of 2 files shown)


wonderflow:
What benefits from adding this dot?

ryanzhang-oss:
> What benefits from adding this dot?

Co-ask, I wonder what's the convention in k8s to specify the path?

hongchaodeng:
Having the dot is a consensus we had when merging the design doc in https://github.com/oam-dev/spec/pull/304.
If anyone has questions please create an issue and discuss the change.
We shouldn't, however, diverge from the design doc in details of the spec.

hongchaodeng:
> what's the convention in k8s to specify the path?

This is JSON path syntax.

hongchaodeng:
> What benefits from adding this dot?

I also recall that the leading dot is a standard syntax of JSON path, which kubectl also uses.

wonderflow:
It makes sense especially I see the docs:


> `.` represents child operator

-- [kubectl jsonpath](https://kubernetes.io/docs/reference/kubectl/jsonpath/)

> The simplest jq program is the expression `.`, which takes the input and produces it unchanged as output.

-- [jq tutorial](https://stedolan.github.io/jq/tutorial/)

Issue #342, opened 2020-03-26
[Proposal] - introduction of Deployment


Haishi2016:
(It's been mentioned during other discussions but I want to put up a separate proposal for it)

# Overview

Conceptually, ApplicationConfiguration serves for several purposes - it describes application topology, it describes configurations, and it represents a deployed unit. With the new application concept, which is being tracked in another proposal, we are separating the first two responsibilities. This proposal proposes we separate representation of a deployed unit out of ApplicationConfiguration and formally introduce a **Deployment** (or Application Instance) object.
A **Deployment** object represent a specific instantiation of an application. It captures information that is relevant to a particular deployment. This will be a natural place to describe rolling updates, canary deployments, blue/green deployments etc.

# Spec
 
The simplest Deployment could be as simple as referencing to an application version:

```yaml
apiVersion: core.oam.dev/v1alpha3
kind: Deployment
metadata:
  name: production
spec:
  versions:
     - name: v1
       applicationRef: 
       - name: my-app
          version: v1
```

And the following sample shows how to describe a canary deployment:

```yaml
apiVersion: core.oam.dev/v1alpha3
kind: Deployment
metadata:
  name: canary
spec:
  versions:
     - name: v1
       applicationRef: 
       - name: my-app
          version: v1
    - name: v2
       applicationRef: 
       - name: my-app
          version: v2
  components:
   - website
  traffic:
  - version: v1
     weight: 70%
  - version: v2
     weight: 30%
  policy:
     primary: v1
     secondary: v2
     rollback:
        errors: 10%
        interval: 1m
     promotion:
        step: 40%
        success: 80%
        interval: 1m
```
Note that the above sample assumes ingress/load-balancer being a built-in concept of the new Application object hence is not explicitly specified here. If this is not the case, we'll need to add a reference to an ingress trait. 

# Value
* Formally separates abstract description of an application and an instance of an application
* Rich language for describing complex deployment scenarios on one or multiple versions

# Implications for a control plane implementation
*  The deployed unit of an application is a Deployment instead of an ApplicationConfiguration

ryanzhang-oss:
Thanks, Haishi. We (Alibaba) already used OAM to describe upgrade/rollback.  In our practice, we apply a trait to a component as a descriptive.  Here are my main take away
1.   Component deployment is mostly application operators' concern and decoupled from the logic of the component. Thus, it seems to be a very natural candidate to be described as a Trait.
[...]

Issue #336, opened 2020-03-17
OAM component versioning mechanism


ryanzhang-oss:
An application Configuration author does not have a way to point to a specific version of a component in the current [spec](https://github.com/oam-dev/spec/blob/1.0.0-alpha2/).  The reason is that OAM spec uses a component's name as the only information to reference to it.

This is needed in **any mission-critical production services** that utilize roll-back as the default emergency backup. In this case, they need a way to specify a previous `version` during the rollback. One can work around this issue if they are willing to take the risk and roll-forward their deployment. However, this is not the case in most of the cloud services providers.

Moreover,  a real workload in the OAM v1alpha2 spec is generated by templating a component with values filled in an applicationConfig. Therefore, this two-level composition nature of a workload makes it more difficult to expose the real version of a component. Here, we will add three proposals below that can solve this problem.

negz:
Some mechanical implementation feedback:

* Kubernetes runtimes can't add arbitrary metadata fields. So no Kubernetes runtime could support setting `metadata.version` on `Component`.
* Kubernetes objects (and I believe OAM objects under the current spec) are uniquely identified by API group, kind, and name (within their namespace). It's within our power to change this for OAM, but not for Kubernetes. So we can't actually lookup a named component by version, nor can multiple versions of the same named component exist at one time.

This isn't to say that we shouldn't _have_ versions; just that the initial proposal isn't feasible.

resouer:
My original idea was revision is implicit" in OAM, but it's tricky in ref model (the main reason Matt put a "name+version" example in current spec).

My current thinking is the version could be annotated on `Component` and version data mapping will be recorded by an independent internal object.  Let's say: [ControllerRevision](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#controllerrevision-v1beta2-apps). 

The controller is responsible for tracking and maintaining the `ControllerRevision` list during `Component` rollout/rollback and choose correct object data. Of course in detail, I assume it will probably only translate different revisions into operations to underlying Deployment and let Rollout Trait controller to perform rollout/rollback (didn't think deeply).

@ryanzhang-oss you could check [the code of StatefulSet](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/statefulset/stateful_set_control.go) for `ControllerRevision` usage, should be straightforward.

hongchaodeng:
[...]

Issue #334, opened 2020-03-13
How can a developer represent the application prior to application configuration?


artursouza:
### Issue
A developer defines the components, on the other hand, an application is composed of multiple components (usually microservices). The only place where components are grouped together is at the ApplicationConfiguration, which is supposed to be authored by ops team.
### Proposal
Create a new concept, called "Application" that will be a way to define a set of components (without traits or param values or scopes). Then, ApplicationConfiguration references the Application and provides all required params and (probably) also add traits and scopes. If the ApplicationConfiguration does not specify one of the components because there is no trait or params to be added, the component is still considered as part of the ApplicationConfiguration since it refers the Application.

wonderflow:
Could you give a real case why a developer want to define a set of components (especially **without traits or param values or scopes**)?

In our cloud team, we really feel application configuration not enough, but that's not for developer. Because we have multiple clusters and multiple availability zones, while application configuration can only used in one cluster. So in my view, the new concept **Application** should leave for that multiple cluster case.

artursouza:
@wonderflow  It might be different things. What you are describing (correct me if I am wrong here) is to have some sort of ApplicationConfigurationTemplate, where some params can be defined in an ApplicationConfiguration. Then each cluster would be a different ApplicationConfiguration for the same ApplicationConfigurationTemplate.

My initial proposal is more dev oriented. In the example of a developer to build an application with 4 different components. The application does not work with only 3 of those, must be all 4. So, the developer could define an Application (without traits or scopes) that says these 4 component together defines Application A. This way, when the app operator is creating an ApplicationConfiguration, all the components for the app are automatically inherited. Only traits, scopes and params must be defined.

Haishi2016:
I agree the Application object should be dev oriented. It should provide an abstract modeling langauge for developers to design the topology of their applications. In my mind, the Application object won't be a generic model that can include everything. Instead, it shall be an opinionated model that abstract common construct of a distributed application. For instance, ingress, scaling limitations, network scopes, as well as sites (for multi-site applications) should be built-in attributes of the Application model. Then, at operation time, operations can decide to attach traits to these properties to realize how the developer intention is fulfilled.

This allows us to have a clear separation - all the developer ever needs to touch is the Application object. Application describes the application topology, a responsibility that is currently mixed into ApplicationConfiguration. Once this is separated out, ApplicationConfiguration is purely for configuration overrides, as it should have been.

[...]

PR #343, opened 2020-03-27
component should be mutable in alpha2

hongchaodeng:


hongchaodeng:
cc @wonderflow @ryanzhang-oss

7.application_configuration.md
```diff
@@ -33,7 +33,7 @@ In addition, as an application configuration is released, its _component instanc
 
 To accommodate this definition of a release, an OAM platform SHOULD make the following assumptions:
 
-- A component (as defined in the [component model](4.component.md) section) is immutable. It should not be changed after creation.
+- A component (as defined in the [component model](4.component.md) section) is mutable. Upgrade can happen by updating the component, e.g. updating image.
```
resouer: Should merge this paragraph with the paragraph below, because change on a Component will results a new release.

hongchaodeng: Sure. Will do it.

negz:
If we mutate a component any existing workload that uses that component will be updated. Is that really what we want? I don't have a strong opinion either way, but it would be very helpful to explain why this change is being made in the PR.

resouer:
@negz What will be the `Component` upgrade story (let's say upgrade image) in Crossplane? 

My current thought is yes, update Component should trigger rollout of the application by following strategy defined in rollout Trait (if defined).

hongchaodeng:
@negz 

In alpha2, the Component is now mutable. This is a fact and the previous message is outdated.

Rollout is a different story from whether Component is mutable or not. It will involve versions, traffic, rollout strategy, etc. Please get involved in https://github.com/oam-dev/spec/issues/342 and discuss Crossplane use cases there.

wonderflow:
In the view of OAM runtime builder, if component is mutable, then component change should trigger AppConfig Controller to reconcile the application(how to reconcile/rollout is another story).  This may add complexity to OAM runtime, but I think it could work.

In this case we may need to add component version automatically in runtime, like revision in K8s object.

resouer:
Since this issue is closely related to https://github.com/oam-dev/spec/issues/336, and https://github.com/oam-dev/spec/issues/334

We will try to bring up a complete story (upgrade, versioning and immutable) for review this week.

ryanzhang-oss:
Please take a look at https://github.com/oam-dev/spec/issues/336 which is very much related to this.

resouer:
LGTM

PR #345, opened 2020-03-28
Don't require a leading dot for JSON field paths

negz:
https://github.com/kubernetes/community/blob/ddc11a7/contributors/devel/sig-architecture/api-conventions.md#selecting-fields

This allows us to leverage existing libraries and patterns designed for Kubernetes field paths, per https://github.com/crossplane/crossplane-runtime/issues/136.

### DIFF
--- a/4.component.md
+++ b/4.component.md
@@ -49,7 +49,7 @@ The Parameters section defines all of the configurable parameters for this compo
 
 Parameter `name` fields must be Unicode letter and number characters. Application Configurations will specify parameters values using this name.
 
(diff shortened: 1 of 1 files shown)


hongchaodeng:
Ref https://github.com/oam-dev/spec/pull/335

If this is a written k8s convention, we should follow and add a line in the spec. Also need to update examples and design doc.

negz:
Ping. Does this seem uncontroversial enough to merge?

resouer:
/lgtm

This one is good enough to merge. Thorough fix could be done by follow up PRs.

hongchaodeng:
@hasheddan 
Note that this is NOT merged into master branch.
Just combine this commit in your follow up PRs.

PR #348, opened 2020-04-10
Update governance model of OAM

resouer:
Update the governance model of OAM with "steering committee + project maintainers" model.

Note that the steering committee will act as maintainers of `oam/spec` at the same time.

### DIFF
--- a/governance.md
+++ b/governance.md
@@ -1,6 +1,6 @@
 # Governance
 
-This document defines project governance for the project.
+This document defines project governance for the projects under oam-dev GitHub organization.
 
 
 ## Steering Committee
@@ -25,6 +25,7 @@ Steering committee is **organization** based.
 - Each organization can have multiple members in the steering committee but only **one** vote.
 - If a steering committee member changes company, his/her membership can not carry on and must be replaced by someone else.
 - All steering committee members must meet following requirements.
+- The organization affiliation of each steering committee member will be recored in OWNERS.md
 
 Requirements:
 
@@ -72,7 +73,7 @@ To become a maintainer of [OAM specification](https://github.com/oam-dev/spec):
 To become a maintainer of other projects under `oam-dev` organization:
 
 - Active contributions for at least 1 month.
-- Get at least 10 non-trivial pull requests merged to the codebase.
+- Get at least 5 non-trivial pull requests merged to the codebase.
 - Primary reviewer for at least 3 non-trivial pull requests from others.
 - Done through PR to update the project’s OWNERS file.
 - Self-nominate or be nominated by existing maintainers in the project.
@@ -85,11 +86,11 @@ If a maintainer is no longer interested or cannot perform the maintainer duties 
 
 ### Github Project Administration
 
-Maintainers will be added to the OAM GitHub organization (if they are not already) and added to the GitHub Maintainers team.
+Maintainers will be added to the OAM GitHub organization (if they are not already) and added to corresponding GitHub maintainers team. The maintainers team is per project.
 
 ## Voting
 
-In general, we prefer that technical issues be resolved at project level by project maintainers. But if a dispute cannot be decided, the steering committee can be called in to decide an issue. If the steering committee members themselves cannot decide an issue, the issue will be resolved by voting. 
+In general, we prefer that technical issues be resolved at project level by project maintainers with "non-objection and negotiation" approach. But if a essential dispute cannot be decided, the steering committee can be called in to decide an issue. If the steering committee members themselves cannot decide an issue, the issue will be resolved by voting. 
 
 - All changes in governance, additions and removals of steering committee members require a **2/3 majority**, while other decisions and changes require only a simple majority.
 - Only steering committee members can vote.
--- a/governance.md
+++ b/governance.md
@@ -1,15 +1,103 @@
 # Governance
 
-## Project Maintainers
-[Project maintainers](OWNERS.md) are responsible for activities around maintaining and updating the Open Application Model (OAM) specification. Final decisions on the spec reside with the project maintainers.
+This document defines project governance for the project.
 
-Maintainers MUST remain active. If they are unresponsive for >3 months, they will be automatically removed unless a [super-majority](https://en.wikipedia.org/wiki/Supermajority#Two-thirds_vote) of the other project maintainers agrees to extend the period to be greater than 3 months.
 
-New maintainers can be added to the project by a [super-majority](https://en.wikipedia.org/wiki/Supermajority#Two-thirds_vote) vote of the existing maintainers. A potential maintainer may be nominated by an existing maintainer. A vote is conducted in private between the current maintainers over the course of a one week voting period. At the end of the week, votes are counted and a pull request is made on the repo adding the new maintainer to the [OWNERS](OWNERS.md) file.
+## Steering Committee
 
-A maintainer may step down by submitting an [issue](https://github.com/oam-dev/spec/issues) stating their intent.
+Steering committee members are responsible for the long term success and adoption of the Open Application Model (OAM) through technical vision and strong commitment to both the projects and community.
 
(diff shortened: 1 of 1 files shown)


wonderflow:
Cool 👍

governance.md
```diff
@@ -1,15 +1,103 @@
...
+- With no objections from existing maintainers in the project.
+
+### Removing a maintainer
+
+If a maintainer is no longer interested or cannot perform the maintainer duties listed above, they should volunteer to be moved to emeritus status. Once volunteered, it will automatically take effect unless objection is raised from existing maintainers. In extreme cases removing maintainers can also occur by a vote of steering committee.
+
+
+### Github Project Administration
+
+Maintainers will be added to the OAM GitHub organization (if they are not already) and added to the GitHub Maintainers team.
```
jbw976, line 88: Is the maintainers team at the organization level? it seems like team usage here may not work well since each project can/will have different maintainers, so having 1 maintainers team at the org level wouldn't be granular enough.  Perhaps adding members directly to each repo's `/settings/access` page may work better? Or having a team per project? 

What level of permission do they get, "maintain"? (i think that's a fairly new rule that github allow now)

governance.md
```diff
@@ -1,15 +1,103 @@
...
+
+## Voting
+
+In general, we prefer that technical issues be resolved at project level by project maintainers. But if a dispute cannot be decided, the steering committee can be called in to decide an issue. If the steering committee members themselves cannot decide an issue, the issue will be resolved by voting.
```
jbw976, line 92: Should project maintainers have a voting process as well that is only scoped to their project, instead of bringing in the steering committee for every conflict for every repo?

governance.md
```diff
@@ -1,15 +1,103 @@
...
+### Becoming a maintainer
+
+To become a maintainer of [OAM specification](https://github.com/oam-dev/spec): 
+- OAM specification is maintained by the OAM steering committee members **only**. Please check the [requirements of becoming steering committee member](#Becoming-a-Steering-Committee-Member).
+
+
+To become a maintainer of other projects under `oam-dev` organization:
+
+- Active contributions for at least 1 month.
+- Get at least 10 non-trivial pull requests merged to the codebase.
```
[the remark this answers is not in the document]
resouer: Agree, I will lower the bar considering projects under oam-dev are mostly building blocks or spec.

governance.md
resouer: Maintainers are per project, so I believe "a maintainer team per project" is the practice. I will elaborate this in this section.

governance.md
resouer: The projects under `oam-dev` are Spec + tools/helpers/tests. 

Since spec is already maintained by Steering Committee. I'd expect other issues in tools/helpers/tests can be solved within their own scopes thru daily maintenance practices such as "non-objection and negotiation" approach. I'd like to keep process in non-spec projects light-weighted and project maintainers there can have their own decision making process for majority cases. WDYT?

governance.md
```diff
@@ -1,15 +1,103 @@
...
+* Accepting feedback from community and mapping to OAM specification.
+* Participating in the conflict resolution and employing the "[organization voting](#Voting)" process when necessary.
+* Actively participate in the regularly scheduled community meetings.
+* Defining common practices and guidelines to OAM implementations in the community.
+
+The current list of steering committee members is published and updated in [OWNERS.md](OWNERS.md#steering-committee).
+
+### Becoming a Steering Committee Member
+
+Steering committee is **organization** based.
```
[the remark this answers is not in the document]
resouer: Yes, good suggestion, the format of OWNERS file will need to be updated.

governance.md
```diff
@@ -1,15 +1,103 @@
 # Governance
 
-## Project Maintainers
-[Project maintainers](OWNERS.md) are responsible for activities around maintaining and updating the Open Application Model (OAM) specification. Final decisions on the spec reside with the project maintainers.
+This document defines project governance for the project.
```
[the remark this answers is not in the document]
resouer: This is for projects under oam-dev organization. Will elaborate.

governance.md
```diff
@@ -1,15 +1,103 @@
...
+
+If you meet above requirements and are interested in joining the steering committee, please update [OWNERS.md](OWNERS.md#steering-committee) by a pull request as self-nomination. Existing steering committee members can also nominate anyone following the same process. Once qualified, steering committee will call for a [vote](#Voting).
+
+### Removing a steering committee member
+
+If a steering committee member is no longer interested or cannot perform the duties listed above, they should volunteer to be moved to emeritus status. Once volunteered, steering committee will call for a [vote](#Voting). In extreme cases, removing existing member can also occur by a vote of the steering committee.
+
+### Nonaffiliated individual
+
+Individuals not associated with or employed by a company or organization are allowed as members of steering committee while this is considered as **rare case**. Clarification will be required in the nomination PR to explain its necessity and must be verified and approved by steering committee.
```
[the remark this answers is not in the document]
resouer: Yes, it's explained in Voting section below, and worth to mention that NI is a rare case in general.

governance.md
jbw976: That sounds fine to me @resouer 👍

resouer:
@jbw976 Just addressed all comments, pleas take a final pass.

governance.md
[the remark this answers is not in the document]
jbw976: Do you expect the owners file to be updated within this PR @resouer?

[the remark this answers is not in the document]
resouer: Hey, I'd expect it as a follow up PR. Note that I've added a sentence here to clarify its format: https://github.com/oam-dev/spec/pull/348/files#diff-898e7f1ceacb493c024554f5a7c87bdfR28

resouer:
@vturecek This proposal has been LGTM by multiple interest parties for long time, while you mentioned you will propose a new process. Could you please send out the new proposal and call for review? Or simply add the new process in this one?

Otherwise I guess we should merge this proposal and iterate it in the future - we can't stay w/o the formal governance doc.

governance.md
```diff
@@ -1,15 +1,104 @@
...
+* Actively participate in the regularly scheduled community meetings.
+* Defining common practices and guidelines to OAM implementations in the community.
+
+The current list of steering committee members is published and updated in [OWNERS.md](OWNERS.md#steering-committee).
+
+### Becoming a Steering Committee Member
+
+Steering committee is **organization** based.
+- Each organization can have multiple members in the steering committee but only **one** vote.
+- If a steering committee member changes company, his/her membership can not carry on and must be replaced by someone else.
```
artursouza, line 26: nit: s/can not/cannot

governance.md
```diff
@@ -1,15 +1,104 @@
...
+- All steering committee members must meet following requirements.
```
artursouza, line 27: nit: move this to the line below.

governance.md
```diff
@@ -1,15 +1,104 @@
...
+
+- All changes in governance, additions and removals of steering committee members require a **2/3 majority**, while other decisions and changes require only a simple majority.
+- Only steering committee members can vote.
+- Organization voting:
+    - Each organization (regardless of the number of steering committee members associated with or employed by that company/organization) receives **one** organization vote.
+    - [Nonaffiliated Individual](#Nonaffiliated-individual) receives one organization vote. 
+
+For formal votes, a specific statement of what is being voted on should be added to the relevant github issue or PR, and a link to that issue or PR added to the maintainers meeting agenda document. Voters should indicate their yes/no vote on that issue or PR, and after a suitable period of time, the votes will be tallied and the outcome noted.
```
artursouza, line 101: All voting, should be recorded in a "decision record" type of file. This way, it can serve as a reference later. Otherwise, the voted decisions might end up being short lived.

resouer:
Sync with @vturecek and he suggested to keep the current simple version of governance doc and he will propose a new one based on https://github.com/CommunitySpecification/1.0.

So close this one.

PR #351, opened 2020-04-26
revise 'component definition' to 'component declaration'

dchjmichael:
I think it's more appropriate to say 'component declaration' here.
I thought it was Component Definition itself.

### DIFF
--- a/7.application_configuration.md
+++ b/7.application_configuration.md
@@ -46,7 +46,7 @@ The [Workload](3.workload.md) chapter of the specification defines how schematic
 
 A component can be deployed into numerous different runtimes. However, at runtime there is a one-to-one correspondence between an _instance of a configuration_ and an _application configuration_.
 
-Conceptually speaking, an application configuration is managed as part of the _application operator_ role. An application configuration includes one or more component instances, each represented by a component definition that defines how an instance of a component spec should be deployed. The component definition has three parts:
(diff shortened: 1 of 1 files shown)


7.application_configuration.md
```diff
@@ -46,7 +46,7 @@ The [Workload](3.workload.md) chapter of the specification defines how schematic
 
 A component can be deployed into numerous different runtimes. However, at runtime there is a one-to-one correspondence between an _instance of a configuration_ and an _application configuration_.
 
-Conceptually speaking, an application configuration is managed as part of the _application operator_ role. An application configuration includes one or more component instances, each represented by a component definition that defines how an instance of a component spec should be deployed. The component definition has three parts:
+Conceptually speaking, an application configuration is managed as part of the _application operator_ role. An application configuration includes one or more component instances, each represented by a component declaration that defines how an instance of a component spec should be deployed. The component declaration has three parts:
```
resouer, line 49: Good catch! I think we can make it even more simpler, how about: 
```suggestion
Conceptually speaking, an application configuration is managed as part of the _application operator_ role. An application configuration references one or more component instances, each represented by a section that defines how an instance of a component spec should be deployed. In detail, this section has three parts:
```

dchjmichael: Good.Now it looks more clear and no ambiguity.

resouer:
/lgtm

Thanks for contributing!

PR #355, opened 2020-05-09
according to community survey, update APAC community call time

wonderflow:
Now more and more people in China join in OAM community, we did a survey and it indicates people love 7:00pm every Tuesday bettor.

image


wonderflow:
@vturecek @suhuruli Can you please help merge it, so community guys could see this change.

PR #356, opened 2020-05-12
make component muteable and add revisionName

### LINKED ISSUE #350: The Component should mutable
# Mutable Component Model

OAM v1alpha1 made a requirement that developer need to define another ComponentSchematic with different name for update (i.e. ComponentSchematic is **immutable**). We've got many feedbacks from real practices in Alibaba as well as [AWS ECS for OAM](https://github.com/awslabs/amazon-ecs-for-open-application-model#upgrade-and-scale-oam-workloads-with-oam-ecs) that it's not a good experience. We agreed on solving it in v1alpha2 with new Component + Workload design. But today, we are still claiming that [Component is immutable](https://github.com/oam-dev/spec/blob/4af9e65769759c408193445baf99eadd93f3426a/6.application_configuration.md#releases) in v1alpha2 spec.

**Goal:**

Developer should be able to modify the Component object directly to trigger further upgrade of the application. i.e. Component is **mutable**.

**Non-goal:**

Revision design around how to track Component change, this should be addressed in: #336

### Change 1

Remove the assumption that a Component could be shared as "template" across multiple AppConfigs. 

Sharing a mutable object is essentially problematic as it mixed ownership from multiple sides. If I defined a Component with `image:v1`, I will be very shocked to see it can be modified by someone else to `image:v2` later.

### Change 2

The relationship between Component and Workload is **not** *template* and *instance* anymore. 

It's important to highlight that what developer maintained is the **Component** object, it should always represent the desired state of his/her application, and it's an "instance". 

Essentially, Component is the envelope to **carry** Workload with extra fields such as `spec.parameters` (and `spec.polices` in the future).  Thus Component is 1:1 mapping to Workload CR and they can have same name in implementation. 

AppConfig still refer the name of Component which implies its controller always tracks the latest version of Component (i.e. the latest desired state of Workload). 

**Revision consideration**

Revision design is off topic but it worth to mention that revisioning could be fully defer to the implementation layer. 

In detail, the implementation can choose/define its own `Revision` object internally to record revision history of Component. For example, using `ControllerRevision` of Kubernetes to store the Component spec as revision data. 

The cons is if a Trait needs to explicitly refer revisions (e.g. A/B test trait), it will need to refer to the name of that internal `Revision` object. Of course, this can be fixed by defining a `Revision` kind in OAM spec for best portability.

Note that developer submitting a Component object only generates a `Revision` object, it does **not** generate Workload object until AppConfig is submitted (i.e. still lazy instantiate).

#### Alternatives

For the relationship of Component and Workload, there're several other options but seems none of them fully meet our goal.

*Option 1:*

Component is an envelope to **generate** Workload which reflects the concept of revision.

That is to say, Component is 1:N mapping to Workload. Every time a developer modified the `spec.workload` of Component, it will generate a Workload object such as: `frontend-workload-v1` and `frontend-worklod-v2`, they could be directly referenced as reversion names in Trait.

But in this case, upon the developer submitted a Component, the system will have to generate Workload object immediately and this will trigger the Workload controller to response. Otherwise, the operator can't know the name of specific Workload in AppConfig later if he need to refer to a specific revision of Component. This breaks the fundamental workflow unless we enforce convention of Workload names like always end with `-v{number}` suffix. The mind burden is relatively heavy for users though.

Note that there're some PaaS/Serveless projects follow this model, because it can rely on its own Workload controller to determine when to generate the real K8s object (e.g. Deployment) after Workload CR is created. But this does not work for us as Workload controller in OAM could be brought by user.

*Option 2:*

Add a `version` field to Component object, so every time a Component is modified, it generates a new version of Component with same name. AppConfig will need to reference Component by `name` + `version`.

But in Kubernetes resource model which OAM adopted, it's infeasible as you can't have two objects with same name and same G/V/K. Also, it still have same issue of "shared Component".

## Conclusion

The experience of modifying Component name for every update has been proven as unacceptable, and we all agree to fix it. But it does not mean the Component should become a "template" object. In practice, it's impossible to maintain such a shared and user facing object in the system. That experience will be broken.

Hence, this proposal is simple in general, we should claim: 

1. Component is mutable;
2. Component can't be shared across AppConfigs.

Changes needed:

1. OAM spec: only documentation change is needed to reflect mutable Component idea, ref: #343;
2. Crossplane: AppConfig controller will need to watch Component change and response. Component can have same name with Workload (not required but helps to simplify the impl).

wonderflow:
Lots of discussion have been made and it seems we all agree with [Approach 3](https://github.com/oam-dev/spec/issues/350#issuecomment-617556483). 

Recently I'm going to implement this feature, let me create this PR to fix #350 .

This also related with #343.

Signed-off-by: 天元 <jianbo.sjb@alibaba-inc.com>


7.application_configuration.md
```diff
@@ -78,11 +78,17 @@ This section defines the instances of components to create with the application
 
 | Attribute | Type | Required | Default Value | Description |
 |-----------|------|----------|---------------|-------------|
-| `componentName` | `string` | Y | | The name of the component of which to create an instance. |
+| `componentName` | `string` | N | | The name of the component of which to bind application configuration. Using componentName represent the ApplicationConfiguration always track the latest revision. |
```
resouer, line 81: it seems in the implementation, it's `component` not `componentName`?

7.application_configuration.md
```diff
@@ -78,11 +78,17 @@ This section defines the instances of components to create with the application
...
+| `revisionName` | `string` | N | | The name of the component revision of which to bind application configuration. |
 | `parameterValues` | [`[]ParameterValue`](#parameterValue) | N | | Overrides of parameters that are exposed by the application scope type defined in `type`. |
 | `scopes` | [`[]Scope`](#scope) | N | | The scopes to be used in the component. A component joins a scope by referencing it. |
 | `traits` | [`[]Trait`](#trait) | N | | The traits to attach to this component instance. |
 
+One of `componentName` and `revisionName` is required. Everytime when a component is changed, a component revision will auto-generated by OAM implementation and can be viewed by user.
```
resouer, line 87: This section is verbose to me. I'd suggest simply claim "the field X as mutually exclusive with Y" and explain everything in the table.

7.application_configuration.md
wonderflow: it's componentName, I confirmed

7.application_configuration.md
wonderflow: thanks, let me fix it

7.application_configuration.md
```diff
@@ -78,7 +78,8 @@ This section defines the instances of components to create with the application
 
 | Attribute | Type | Required | Default Value | Description |
 |-----------|------|----------|---------------|-------------|
-| `componentName` | `string` | Y | | The name of the component of which to create an instance. |
+| `componentName` | `string` | N | | The name of the component whose latest revision will be bind with application configuration. When the spec of the referenced component changes, ApplicationConfiguration will automatically migrate all trait affect from the prior revision to the new one. This is mutually exclusive with revisionName.|
```
ryanzhang-oss, line 81: why is the componentName not required anymore?

resouer, line 81: Needs better wording.

```suggestion
| `componentName` | `string` | N | | The name of the component of which to create an instance. The revision of this instance will always reflect the latest change of the component (i.e. always track latest revision). This is mutually exclusive with revisionName.|
```

7.application_configuration.md
```diff
@@ -78,7 +78,8 @@ This section defines the instances of components to create with the application
 
 | Attribute | Type | Required | Default Value | Description |
 |-----------|------|----------|---------------|-------------|
-| `componentName` | `string` | Y | | The name of the component of which to create an instance. |
+| `componentName` | `string` | N | | The name of the component of which to create an instance. The revision of this instance will always reflect the latest change of the component (i.e. always track latest revision). This is mutually exclusive with revisionName.|
+| `revisionName` | `string` | N | | RevisionName of a specific component revision to which to bind application configuration. This is mutually exclusive with componentName. |
```
resouer, line 82: Sorry, better wording maybe.

```suggestion
| `revisionName` | `string` | N | | The name of specific component revision of which to create an instance, i.e. always stick to specific revision until this field is updated. This is mutually exclusive with componentName. |
```

resouer:
After this is implemented, we could add a section about why OAM choose to embed workload CR in component instead of ref workload CR - revision is the main concern here.

Issue #402, opened 2020-09-09
Proposal: Data Placement Policy


chilung:
Suppose an edge service in Bay Area serve me well and then I fly to New York, another edge service serve me well too. However, the edge service in New York does not know or have my data in Bay Area edge location. A Cloud service does not have this kind of issue. Does or how OAM addresses this development requirement?
It depends on the policy on (1) auto data migration or (2) connect to data nearest to the edge service automatically (data is already there, just need to let the edge service know) according to latency requirements and data amount to be migrated.
I believe it can be an important topic to realize edge computing.

We propose a mechanism to define the data placement policy: In Cloud, in Edge Read Only, in Edge Read Write Non-Shareable, and in Edge Read Write Shareable.
1. For those data defined 'In Cloud', the data will stay in Cloud and will not migrate to the Edge where is near users.
2. For those data defined 'in Edge Read Only', the data will have an instance in Edge near users. It is possible to have multiple instances for the same data entity spread among multiple edge sites. We need to define the behavior of data migration according to storage usage and data expiration policy.
3. For those data defined 'in Edge Read Write Non-Shareable', the data will have an instance in Edge near users. It is also possible to have multiple instances for the same data entity spread among multiple edge sites. We need to define the behavior of data migration according to storage usage and data expiration policy. It will be further defined whether these multiple instances should be synchronized back and saved to Cloud or not.
4. For those data defined 'in Edge Read Write Shareable', although the data also has multiple instances among multiple edge sites, state should be maintained consistence among these instances.

In Open Application Model, we can define it as 'DataPlacementPolicy' scope. User can bind DataPlacementPolicy to different Components in his application. The implement of DataPlacementPolicy will handle the data migration detail based on the Scope kind.

resouer:
@chilung Based on the discussion on gitter, I belive this is in scope of OAM and a valuable use case. I'd suggest we start from including this ApplicationScope from [extend category](https://github.com/oam-dev/spec/blob/master/5.application_scopes.md#categories-of-application-scope) so we can begin to iterate its spec.

Would you like to draft the first version of spec like `HealthScope`: https://github.com/oam-dev/spec/blob/master/standard/scopes/health_scope.md

It could be:

```yaml
apiVersion: extend.oam.dev/v1alpha2
kind: DataPlacementScope
spec:
  ...
```

PR #404, opened 2020-09-14
Data Placement Policy scope

chilung:
in responding to issue #402, draft the first version of data placement scope spec.


extend/data_placement_scope.md
```diff
@@ -0,0 +1,34 @@
...
+
+For those data defined in Edge, read/write and exclusive, the data will have an instance in Edge near users. It is also possible to have multiple instances for the same data entity spread among multiple edge sites. We need to define the behavior of data migration according to storage usage and data expiration policy. It will be further defined whether these multiple instances should be synchronized back and saved to Cloud or not.
+
+For those data defined in Edge, read only, the data will have an instance in Edge near users if exclusive. It is also possible to have multiple instances for the same data entity spread among multiple edge sites if shared. We need to define the behavior of data migration according to storage usage and data expiration policy.
+
+In Open Application Model, we can define it as 'DataPlacement' scope. Users can bind 'DataPlacement' to different Components in their application. The implementation of DataPlacemento will handle the data migration detail based on the Scope kind.
```
hongchaodeng, line 13:
```suggestion
In Open Application Model, we can define it as 'DataPlacement' scope. Users can bind 'DataPlacement' to different Components in their application. The implementation of DataPlacement will handle the data migration detail based on the Scope kind.
```

hongchaodeng (commented):
I'm wondering how a user would use this scope? Could you add a section for the user scenario?

extend/data_placement_scope.md
chilung: I will add a section for scenario and also fix the typo.

chilung: I have added a section for scenario and also fixed the typo.
I also have modified some descriptions, such as 'in-cloud' to replace in Cloud, etc, to keep the explanation and definition in consistence.

extend/data_placement_scope.md
```diff
@@ -0,0 +1,38 @@
...
+
+In Open Application Model, we can define it as 'DataPlacement' scope. Users can bind 'DataPlacement' to different Components in their application. The implementation of DataPlacement will handle the data migration detail based on the Scope kind.
+
+## Sample User Scenario
+
+In mixed reality application, users usually touch and manipulate the holograms of targeted objects rendered in the vision of real world. The digital twins of the targeted objects can be specified as components bound with 'in-edge' attribute of 'DataPlacement' scope in consideration of the requirement of low latency. The template part of digital twins can be further specified with 'RO' attribute to reflect the fact of templates cannot be modified. On the other hand, the configuration part of digital twins can be specified with 'RW' and 'exclusive' in single user mode. In multi-user mode, The configuration part of digital twins can be specified with 'RW' and 'shared'.
```
hongchaodeng, line 17: To be honest, I don't understand this scenario. Could you provide a simpler case?

Additionally, I'm also expecting you to go through the scenario with the spec, not just describing the use case.

hongchaodeng (approved):
LGTM

I will wait a day or two in case other reviewers might have more opinions.

samiarahman:
I am joining this community from data engineering point of view, the data placement scope that you speak of would only apply to components that persist or serve data, how do we ensure the scope does not get forexample applied to the front end component etc.

chilung:
> I am joining this community from data engineering point of view, the data placement scope that you speak of would only apply to components that persist or serve data, how do we ensure the scope does not get forexample applied to the front end component etc.

Could you elaborate this issue in more detail?

resouer:
@chilung For example, for traits we have a `appliestTo` field to say whether this trait applies to certain components or not. This would solve the concern of @samiarahman maybe?

samiarahman:
@resouer  yes that makes sense. Sorry for the late reply.

resouer:
Cool, I will merge this design as a new extended Scope, thanks for contribution!

Issue #411, opened 2020-10-06
Workload settings and workload definitions in v0.2.2 working draft


resouer:
> Claimer: this issue is discovered by Brian Bialecki from ACR and reported during OAM community meeting.

1. The workload settings and workload CR should be manually exclusive.
2. Workload definitions should use real workload types instead of made up types in the core workload examples.

PR #413, opened 2020-10-06
Fix mutually exclusive fields and definition sample

resouer:
Fixes: #411 

1. The `workload` section should be mutually exclusive with `type` and `name`.
2. The `WorkloadDefinition` sample should use real workload types instead of fake workload types.

### DIFF
--- a/3.component.md
+++ b/3.component.md
@@ -29,10 +29,10 @@ The spec defines the constituent parts of a component.
 
 | Attribute | Type | Required | Default Value | Description |
 |-----------|------|----------|---------------|-------------|
-| `type` | string | Y | | The type of workload|
-| `settings` | [`WorkloadSettings`](#workloadsettings) | Y | | The workload-specific configuration. |
-| `parameters` | [`[]Parameter`](#parameter) | N | | The component's configuration options. |
-| `workload` | [`Workload`](#workload) | Y | | <kbd>LEGACY</kbd> The specification of workload that should be passed to the runtime. |
+| `type` | string | N | | The type of workload. |
+| `settings` | [`WorkloadSettings`](#workloadsettings) | N | | The workload-specific configuration. |
+| `parameters` | [`[]Parameter`](#parameter) | N | | The component's configurable parameters. |
+| `workload` | [`Workload`](#workload) | N | | <kbd>LEGACY</kbd> The full specification of workload that should be passed to the runtime. This field is mutually exclusive with `type` and `settings`.|
 
 #### Type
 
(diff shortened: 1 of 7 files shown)


3.component.md
```diff
@@ -73,10 +73,10 @@ metadata:
   annotations:
     version: v1.0.0
     description: >
-      Sample component that describes the administrative interface for our Twitter bot.
+      Sample component that describes a very cool workload.
 spec:
-  type: Server
-  settings:
+  type: Server # <--- the type of the workload
+  settings: # <---- the settings of the workload
```
wonderflow, line 79: This example seems to be different with current implementation, right?

resouer: Yes, this change targets for OAM spec v0.2.2 release, and what we have implemented is OAM spec v0.2.1. The change has been synced with MSFT in maintainer meeting. We could sync internally if needed.

ryanzhang-oss: The current crossplane implementation doesn't have the `type` and `settings` but I can see that the example format here is less K8s style and more generic.

4.workload_definitions.md
```diff
@@ -50,7 +50,7 @@ The specification section contains a reference to the workload definition.
 | Attribute | Type | Required | Default Value | Description |
 |-----------|------|----------|---------------|-------------|
 | `definitionRef` | [`DefinitionRef`](#DefinitionRef) | Y | | Workload schema reference. |
-| `extension` | unstructured | N | | <kbd>EXPERIMENTAL</kbd> A unstructured field for the implementation to reuse definition object carry extra information for UI or helper purpose. For example, a markdown description of how to use this workload etc. The implementation should NOT abuse this field for functional purpose. |
+| `extension` | unstructured | N | | <kbd>EXPERIMENTAL</kbd> A unstructured field for the implementation to reuse definition object carry extra information for UI or helper purpose. For example, a markdown description of how to use this workload etc. The implementation should NOT abuse this field for other purpose. |
```
ryanzhang-oss, line 53:
```suggestion
| `extension` | unstructured | N | | <kbd>EXPERIMENTAL</kbd> An unstructured field for the OAM platform builders to carry extra information in the definition objects for UI or helper purposes. For example, this field can contain a markdown description of how to use the workload. The implementation should NOT abuse this field for other purposes. |
```

3.component.md
resouer: @ryanzhang-oss Yes, it's based on the discussion in maintainer call w/ MSFT folks, and it's easier to implement with compatibility than our previous discussion. OAM k8s runtime could implement it after spec v0.2.2 (i.e. the current working draft of this PR) is released.

wonderflow (approved):

3.component.md
captainroy-hy, line 79: the `fieldPaths` of `parameters` used in the sample may be revised according to new spec field structure.
> "spec.containers[0].image" ==> "spec.settings.containers[0].image"

resouer: Good catch, just fixed.

hongchaodeng (approved):
lgtm

